PR #157, opened 2016-10-07
Fix pending deployments

maciejmrowiec:


maciejmrowiec:
@GregorioDiStefano @pasinskim @mchalski

PR #159, opened 2016-10-11
Introduce logging

mchalski:
Issues: MEN-533

Doesn't include propagating the requestid to inventory (yet).

@maciejmrowiec @kjaskiewiczz 

EDIT: this still needs some fixes in tests/mocks, I'll keep them coming
EDIT2: all fixes done


middleware.go
```diff
@@ -57,11 +61,14 @@ var DefaultDevStack = []rest.Middleware{
...
 	&rest.JsonIndentMiddleware{},
+	&requestid.RequestIdMiddleware{},
 }
 
 var DefaultProdStack = []rest.Middleware{
 
 	// logging
+	&requestlog.RequestLogMiddleware{},
+	&accesslog.AccessLogMiddleware{Format: accesslog.SimpleLogFormat},
 	&rest.AccessLogJsonMiddleware{
```
marekswiecznik: should be removed, duplicates AccessLogMiddleware, but in json format (already removed in deviceauth, adm and inventory)

mchalski: ahh, yes, thx

middleware.go
```diff
@@ -24,6 +24,9 @@ import (
 
 	"github.com/ant0ine/go-json-rest/rest"
 	"github.com/mendersoftware/deployments/config"
+	"github.com/mendersoftware/deployments/utils/accesslog"
```
mchalski: up to @maciejmrowiec 

it's not clear to me what's the status of our shared library. I remember asking a couple weeks ago if there's a repo for it, and found out that we're not doing this after all? can't recall if it was jira/github/slack.

middleware.go
mchalski: fixed

resources/deployments/controller/controller_deployments_external_test.go
```diff
@@ -44,6 +46,16 @@ func makeDeviceAuthHeader(claim string) string {
 		base64.StdEncoding.EncodeToString([]byte(claim)))
 }
 
+func makeApi(router rest.App) *rest.Api {
+	api := rest.NewApi()
+	api.Use(
+		&requestlog.RequestLogMiddleware{},
```
maciejmrowiec: something I noticed on other services, will this throw lot of spam to stderr while executing tests? if so, maybe it would be good to redirect logs to `/dev/null`

resources/images/view/view.go
```diff
@@ -44,12 +45,13 @@ func (p *RESTView) RenderSuccessGet(w rest.ResponseWriter, object interface{}) {
 	w.WriteJson(object)
 }
 
-func (p *RESTView) RenderError(w rest.ResponseWriter, err error, status int) {
+func (p *RESTView) RenderError(w rest.ResponseWriter, err error, status int, l *log.Logger) {
+	l.F(log.Ctx{log.LogHttpCode: status}).Error(err.Error())
```
maciejmrowiec: Access log middleware already logs http codes

mchalski: ok, fixed

resources/deployments/controller/controller_deployments_external_test.go
mchalski: ok, but how would you do it?

middleware.go
mchalski: ok, but let's handle it in a separate PR, current TODOs are:
1. fixes for comments so far
2. reqid handling (a lot is done, but still in progress)
3. refactoring to lib

2 and 3 are orthogonal, so I suggest we close this PR after 1 is handled to avoid messy converstations on unrelated subjects.

resources/deployments/controller/controller_deployments_external_test.go
bboozzoo: I think we can take that as an enhancement and fix it later. How does that sound?

maciejmrowiec: why not do it straight away? later usually means never

bboozzoo: As I understood there's some planned refactoring for transitioning to https://github.com/mendersoftware/go-lib-micro Now that this issue of a custom logger has been raised, it makes sense to add that functionality in the lib (to the benefit of all) and just update deployments when pulling in the lib.

mchalski: ok, your call. I need to refactor the middleware then:
- right now it derives from a global logger
- I'll make it accept a base logger, which will override the global one if the user provides it

then the user will be able to pass the root/reference logger with whatever config they please.

I just need an ACK that we're on the same page before I start.

maciejmrowiec: Makes sense.

mchalski: ok, I stacked 2 additional commits related to this, to make this easier to read

maciejmrowiec:
looks good, shall we ?

PR #160, opened 2016-10-12
Unify API error output

marekswiecznik:
Issues: MEN-501

Basicaly, all 500 responses were changed to they return `{"error": "internal error", "request_id": "..."}` and log error from lower levels. 4xx errors return mostly general errors, in case of validation - more detailed.

Additionaly all errors include request id in response body.

@maciejmrowiec  @mchalski

### DIFF
--- a/resources/images/view/view.go
+++ b/resources/images/view/view.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/ant0ine/go-json-rest/rest"
 	"github.com/mendersoftware/deployments/utils/log"
+	"github.com/mendersoftware/deployments/utils/requestid"
 )
 
 // Headers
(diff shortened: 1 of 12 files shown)


marekswiecznik:
added request id to error response body (new commits)

resources/images/view/view.go
```diff
@@ -50,6 +50,11 @@ func (p *RESTView) RenderError(w rest.ResponseWriter, err error, status int, l *
 	rest.Error(w, err.Error(), status)
 }
 
+func (p *RESTView) RenderInternalError(w rest.ResponseWriter, err error, l *log.Logger) {
+	l.F(log.Ctx{log.LogHttpCode: http.StatusInternalServerError}).Error(err.Error())
```
mchalski: remove http code, logged by requestlog middleware

mchalski:
1 small comment, otherwise +1

resources/images/view/view.go
marekswiecznik: removed in next commit: https://github.com/mendersoftware/deployments/pull/160/commits/d69e1312d6595813516f5fd0679c68cbe8633beb#diff-257cd2d9cfbaa0f72550558a1a99ba25L53

PR #161, opened 2016-10-13
Image upload fixed

kjaskiewiczz:
There was some issues with old version of aws-sdk-go library (wrong handling of ReadSeeker) so I've updated aws-sdk-go library to latest tagged version (v1.4.15).
In this new version of aws-sdk-go library there are still some issues, but I've introduced workaround for them (as suggested by amazon).
I've tested image upload with fakes3, minio.io and amazon s3 and it looks like it's working now.
I'm not sure if vendoring is done properly, so please take a look at this.

@maciejmrowiec @bboozzoo @mchalski @marekswiecznik


maciejmrowiec:
by browser does not play well with displaying the update...
but patch itself looks good

PR #165, opened 2016-10-19
Introduce already-installed device deployment status

marekswiecznik:
Issues: MEN-710

@maciejmrowiec @bboozzoo


docs/integrations_api.yml
```diff
@@ -486,6 +487,9 @@ definitions:
       noimage:
         type: integer
         description: Do not have apropriate image for device type.
+      already-installed:
+        type: integer
+        description: Number of deployments devices being already installed.
```
maciejmrowiec: Let's make it slightly more specific - may be slightly confusing how it's different from success.

`Number of devices unaffected by upgrade, since they are already running the specified software version.`

marekswiecznik: ok, fixed.

PR #166, opened 2016-10-20
Minio server supported.

kjaskiewiczz:
Minio accepts only "us-east-1" region.
Bucket must be created explicitly.

### DIFF
--- a/resources/images/s3/filestorage.go
+++ b/resources/images/s3/filestorage.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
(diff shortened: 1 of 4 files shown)


resources/images/s3/filestorage.go
```diff
@@ -54,10 +54,21 @@ func NewSimpleStorageServiceStatic(bucket, key, secret, region, token, uri strin
...
 
+	client := s3.New(sess)
+
+	// minio requires explicit bucket creation
+	cparams := &s3.CreateBucketInput{
+		Bucket: aws.String(bucket), // Required
+	}
+
+	// Create a new bucket using the CreateBucket call.
+	client.CreateBucket(cparams)
```
maciejmrowiec: what if the bucket already exist or w fail to create one?

kjaskiewiczz: I'm not sure about error handling here.
If we cannot create bucket we should probably exit (panic?).
If the bucket already exists it is probably ok;
maybe we should check if this bucket exists and try to create it only if not,
or maybe we shoud check if there is a way to setup minio server with predefined buckets

maciejmrowiec: erring up I think in case of issues creating bucket is OK, we do that in some other cases 

staring service with bucket already existing is going to be pretty common case:
* starting second, third etc instance in setup
* manually setup bucket in S3
* service restarts

create if not exists sounds like a good idea, we don't need to require it as precondition

PR #167, opened 2016-10-25
Aborted dev deployment status

marekswiecznik:
Issues: MEN-717

I also fixed a bug in `TestDeploymentModelUpdateDeviceDeploymentStatus` (discovered when adding test case with `aborted` status).

@kjaskiewiczz @maciejmrowiec


resources/deployments/deployment_external_test.go
```diff
@@ -292,6 +294,13 @@ func TestDeploymentGetStatus(t *testing.T) {
 			},
 			OutputStatus: "finished",
 		},
+		"Failed + Aborted": {
+			Stats: map[string]int{
+				DeviceDeploymentStatusFailure:     1,
```
kjaskiewiczz: it's gofmt ok with this?

marekswiecznik: nope, corrected, thx

kjaskiewiczz:
+1

maciejmrowiec:
What about update to devices_api docs?
Not sure if I missed but I don't see mechanism that forbids changing aborted status into something else.

marekswiecznik:
Right, I missed that, thx @maciejmrowiec

marekswiecznik:
Added mechanism to forbid setting device deployment status when it is already aborted. @maciejmrowiec

PR #168, opened 2016-10-26
Assign MIME type for artifact

marekswiecznik:
Issues: MEN-676

Theoretically it should be enough to set `Content-Type` in `PutObjectInfo`. For Amazon S3 it is, but for minio - it is not. I tried to trace down this bug, but it's very hard to do so, because I can't even verify if metadata are stored correctly. There is a workaround for that - setting response content-type when generating download link and it is exactly what I did.

@kjaskiewiczz  @bboozzoo

### DIFF
--- a/resources/deployments/model/deployments.go
+++ b/resources/deployments/model/deployments.go
@@ -34,21 +34,26 @@ type DeploymentsModel struct {
 	deviceDeploymentLogsStorage DeviceDeploymentLogsStorage
 	imageLinker                 GetRequester
 	deviceDeploymentGenerator   Generator
+	imageContentType            string
 }
 
-func NewDeploymentModel(
-	deploymentsStorage DeploymentsStorage,
-	deviceDeploymentGenerator Generator,
-	deviceDeploymentsStorage DeviceDeploymentStorage,
-	deviceDeploymentLogsStorage DeviceDeploymentLogsStorage,
-	imageLinker GetRequester,
-) *DeploymentsModel {
+type DeploymentsModelConfig struct {
+	DeploymentsStorage          DeploymentsStorage
+	DeviceDeploymentsStorage    DeviceDeploymentStorage
+	DeviceDeploymentLogsStorage DeviceDeploymentLogsStorage
+	ImageLinker                 GetRequester
+	DeviceDeploymentGenerator   Generator
+	ImageContentType            string
+}
+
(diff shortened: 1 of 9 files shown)


kjaskiewiczz:
minio/cmd/fs-v1.go
453     // Save additional metadata only if extended headers such as "X-Amz-Meta-" are set.
content-type is in the "additional metadata"...

I think this is why it doesn't work with minio.
It is also possible that minio is implementing amazon API correctly, but amazon is not (I've seen such situation before).

marekswiecznik:
@kjaskiewiczz Unfortunately that is not the case here. If you set 'Content-Type' in metadata, then when downloading you will get such headers:
`Content-Type:		application/octet-stream`
`X-Amz-Meta-Content-Type:		application/vnd.mender-artifact`

resources/deployments/model/deployments.go
```diff
@@ -34,6 +34,7 @@ type DeploymentsModel struct {
 	deviceDeploymentLogsStorage DeviceDeploymentLogsStorage
 	imageLinker                 GetRequester
 	deviceDeploymentGenerator   Generator
+	imageContentType            string
 }
 
 func NewDeploymentModel(
```
bboozzoo: Not a huge problem, but the patch adds a 6th positional parameter to `NewDeploymentModel`. IMO it's a bit too many, even by Go standards. Do you think it could be replaced with a struct? e.g. 
```
type DeploymentModelConfig struct {
 	deploymentsStorage DeploymentsStorage
 	deviceDeploymentGenerator Generator
 	deviceDeploymentsStorage DeviceDeploymentStorage
 	deviceDeploymentLogsStorage DeviceDeploymentLogsStorage
 	imageLinker GetRequester
	imageContentType string
}

func NewDeploymentModel(config DeploymentModelConfig) {
...
}
```

@maciejmrowiec second opinion?

maciejmrowiec: makes sense, doe how we would handle incorrectly constructed config? return extra error from New ?

maciejmrowiec: in this case wouldn't we need constructor for config validation?

maciejmrowiec: alright, let's do this then

marekswiecznik:
@bboozzoo fixed

PR #170, opened 2016-11-21
Indicate presence of the deployment log for the device.

ciekawsky:
Issue: MEN-746
Signed-off-by: Piotr Bara <piotr.bara@rndity.com>
 @mendersoftware/rndity 

Updating tests currently.

### DIFF
--- a/resources/deployments/model/deployments_external_test.go
+++ b/resources/deployments/model/deployments_external_test.go
@@ -799,9 +799,10 @@ func TestDeploymentModelSaveDeviceDeploymentLog(t *testing.T) {
 		InputDeviceID     string
 		InputLog          []deployments.LogMessage
 
-		InputModelError    error
-		InputHasDeployment bool
-		InputHasModelError error
+		InputModelError     error
+		InputHasDeployment  bool
+		InputHasModelError  error
+		InputUpdateLogError error
 
 		OutputError error
 	}{
@@ -809,10 +810,10 @@ func TestDeploymentModelSaveDeviceDeploymentLog(t *testing.T) {
 			InputDeploymentID:  "f826484e-1157-4109-af21-304e6d711560",
 			InputDeviceID:      "123",
 			InputLog:           messages,
-			InputModelError:    errors.New("storage issue"),
+			InputModelError:    errors.New("Storage issue"),
 			InputHasDeployment: true,
 
-			OutputError: errors.New("storage issue"),
+			OutputError: errors.New("Storage issue"),
 		},
 		{
 			InputDeploymentID:  "ID:234",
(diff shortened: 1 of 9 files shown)


pasinskim:
@ciekawsky looks like this one is breaking the build.

resources/deployments/model/deployments_external_test.go
```diff
@@ -859,6 +859,11 @@ func TestDeploymentModelSaveDeviceDeploymentLog(t *testing.T) {
 			testCase.InputDeploymentID,
 			testCase.InputDeviceID).
 			Return(testCase.InputHasDeployment, testCase.InputHasModelError)
+		deviceDeploymentStorage.On("UpdateDeviceDeploymentLogAvailability",
+			testCase.InputDeviceID,
+			testCase.InputDeploymentID,
+			true).
+			Return(nil)
```
bboozzoo: You're checking success path only right now. How about a scenario in which `UpdateDeviceDeploymentLogAvailability` fails?

resources/deployments/mongo/device_deployments.go
```diff
@@ -198,6 +199,30 @@ func (d *DeviceDeploymentsStorage) UpdateDeviceDeploymentStatus(deviceID string,
...
+
+	session := d.session.Copy()
+	defer session.Close()
+
+	selector := bson.M{
+		StorageKeyDeviceDeploymentDeviceId:     deviceID,
+		StorageKeyDeviceDeploymentDeploymentID: deploymentID,
+	}
+
+	update := bson.M{"$set": bson.M{StorageKeyDeviceDeploymentIsLogAvailable: log}}
```
bboozzoo: Update formatting so that it looks like this:
```
bson.M{
    "$set": bson.M{
        StorageKeyDeviceDeploymentIsLogAvailable: log
     }
}
```

ciekawsky: Done.

resources/deployments/model/deployments_external_test.go
ciekawsky: Thank you, an appropriate case is now added.

resources/deployments/mongo/device_deployments_test.go
```diff
@@ -286,7 +378,6 @@ func TestAggregateDeviceDeploymentByStatus(t *testing.T) {
 		session := db.Session()
 		store := NewDeviceDeploymentsStorage(session)
 
-		// deployments are created with status DeviceDeploymentStatusPending
```
ciekawsky: This was not true. In this test, each DeviceDeployment were specifically created with different statuses to cover all of them

PR #171, opened 2016-11-22
artifact id in request

bboozzoo:

Although [MEN-744](https://tracker.mender.io/browse/MEN-744) mentions that the device is supposed to send both `artifact` and `device_type`, I have not found any use for `device_type` in the API handler. Feel free to point out the reasons why `device_type` is needed.

### DIFF
--- a/docs/devices_api.yml
+++ b/docs/devices_api.yml
@@ -27,7 +27,7 @@ responses:
       $ref: "#/definitions/Error"
 
 paths:
-  /device/update:
+  /device/deployments/next:
     get:
       summary: Get a next update
       description: |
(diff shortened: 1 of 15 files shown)


kacf:
Isn't `device_type` supposed to be matched against `device_types_compatible` in the artifact file, and then the update is offered only if there's a match?

bboozzoo:
Maybe it'll work like this later, when support for _new_ artifacts is added.

As of now, there's just no use of this field. Device deployments (that describes a deployment for each devices) are created upfront when a new deployment is added. This is the only time when device type is looked into.

kacf:
Ok, well at least you know what the intention was from the client side! IIRC, I think the reason we wanted the `device_type` to come with the request, is so that the deployment service doesn't have to ask the inventory. I think @maciejmrowiec wanted to keep deployments and inventory as separate as possible.

But yes, ultimately @maciejmrowiec needs to answer this.

bboozzoo:
ping

PR #172, opened 2016-11-24
artifact_name in next update request

bboozzoo:
@mendersoftware/rndity @kacf @pasinskim @maciejmrowiec @GregorioDiStefano

### DIFF
--- a/resources/deployments/controller/controller_deployments.go
+++ b/resources/deployments/controller/controller_deployments.go
@@ -173,8 +173,12 @@ func (d *DeploymentsController) AbortDeployment(w rest.ResponseWriter, r *rest.R
 	d.view.RenderEmptySuccessResponse(w)
 }
 
+const (
+	GetDeploymentForDeviceQueryArtifact   = "artifact_name"
+	GetDeploymentForDeviceQueryDeviceType = "device_type"
+)
+
 func (d *DeploymentsController) GetDeploymentForDevice(w rest.ResponseWriter, r *rest.Request) {
-	const queryArtifact = "artifact"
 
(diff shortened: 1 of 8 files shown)


resources/deployments/controller/controller_deployments.go
```diff
@@ -173,8 +173,12 @@ func (d *DeploymentsController) AbortDeployment(w rest.ResponseWriter, r *rest.R
 	d.view.RenderEmptySuccessResponse(w)
 }
 
+const (
+	GetDeploymentForDeviceQueryArtifact   = "artifact_name"
+	GetDeploymentForDeviceQueryDeviceType = "device_type"
```
kacf: If you are making it here, why not include the attribute `device_type` also in this PR?

kacf: I mean *use* the attribute, not just define it.

bboozzoo: What do you mean by use? Extract it from url.Values just for the sake of getting it? Because there's no other use for it at this time.

bboozzoo:
Don't merge yet, I'm moving some logic from `controller` to `model`.

Seems like the naming is off, `controller` is in fact a transport encoder/decoder. `model` is not data model but the actual service `controller`. Not that anyone cares, it just looks a bit weird.

resources/deployments/controller/controller_deployments.go
kacf: Hmm, never mind. I thought it would be used to compare against the artifact file that the device type is correct, but I guess that's not implemented yet.

bboozzoo:
Should be ready for review now

PR #178, opened 2016-11-30
support multiple artifact device types when creating deployment

mchalski:
https://tracker.mender.io/browse/MEN-721

`meta_yocto.device_type` becomes `meta_yocto.device_types_compatible`.- that's the only change.

note that mongo doesn't care if it matches a single value or an array of values; the latter has 'select if any element matches' semantics.

 @bboozzoo @kjaskiewiczz @maciejmrowiec

### DIFF
--- a/resources/images/mongo/dbsetup_test.go
+++ b/resources/images/mongo/dbsetup_test.go
@@ -1,0 +1,46 @@
+// Copyright 2016 Mender Software AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


resources/images/mongo/dbsetup_test.go
```diff
@@ -0,0 +1,46 @@
...
+// Overwrites test execution and allows for test database setup
+func TestMain(m *testing.M) {
+
+	dbdir, _ := ioutil.TempDir("", "dbsetup-test")
+	// os.Exit would ignore defers, workaround
+	status := func() int {
+		// Start test database server
+		db = &dbtest.DBServer{}
+		db.SetPath(dbdir)
+		// Tear down databaser server
```
pasinskim: s/databaser/database/

PR #187, opened 2016-12-23
Dockerfile: migrate to alpine:3.4

mchalski:
Issues: MEN-886

Signed-off-by: mchalczynski <marcin.chalczynski@rndity.com>

@mendersoftware/rndity @maciejmrowiec @GregorioDiStefano

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,8 @@
-FROM iron/base
+FROM alpine:3.4
+
+RUN apk update && apk upgrade && \
+     apk add ca-certificates && \
+     rm -rf /var/cache/apk/*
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM iron/base
+FROM alpine:3.4
 
 COPY ./deployments /usr/bin/
```
bboozzoo: We'll need to install ca-certificates package, otherwise using S3 (the real one) will not work.

```
RUN apk update && apk upgrade \
     apk add ca-certificates && \
     rm -rf /var/cache/apk/*
```

mchalski: fixed

PR #188, opened 2016-12-27
Artifact upload handling changed: temoporary file not used anymore

kjaskiewiczz:
Artifact upload handling changed.

Old approach:
- whole artifact file is uploaded to backed and saved in temporary file;
- artifact from temp file is uploaded to the file server.

New approach:
- metadata is read into local buffer;
- local buffer with metadata and rest of the uploaded file are uploaded to file server.

@bboozzoo @maciejmrowiec @mchalski

### DIFF
--- a/resources/images/controller/images_external_test.go
+++ b/resources/images/controller/images_external_test.go
@@ -87,7 +87,7 @@ func (fim *fakeImageModeler) DeleteImage(imageID string) error {
 }
 
 func (fim *fakeImageModeler) CreateImage(
-	imageFile *os.File,
+	artifact io.Reader,
 	metaConstructor *images.SoftwareImageMetaConstructor,
 	metaArtifactConstructor *images.SoftwareImageMetaArtifactConstructor) (string, error) {
 	return "", nil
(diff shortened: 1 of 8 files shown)


resources/images/s3/filestorage.go
```diff
@@ -15,8 +15,10 @@
 package s3
 
 import (
+	//"bufio"
```
bboozzoo: remove

kjaskiewiczz: done

resources/images/s3/filestorage.go
```diff
@@ -15,8 +15,10 @@
...
 	"fmt"
-	"os"
+	"io"
+	//"os"
```
kjaskiewiczz: done

resources/images/controller/images_external_test.go
```diff
@@ -463,7 +463,7 @@ func TestSoftwareImagesControllerNewImage(t *testing.T) {
 
 		model.On(
 			"CreateImage",
-			mock.AnythingOfType("*os.File"),
+			mock.Anything,
```
bboozzoo: I think it should be possible to use `mock.MatchedBy()` here, like this:

```
model.On(
    "CreateImage",
    mock.MatchedBy(func(ir interface{}) bool {
        _, ok := ir.(io.Reader)
        return ok
     }),
    ...
    )
```

care to try?

kjaskiewiczz: looks like it's working
updated

bboozzoo:
@mendersoftware/rndity @maciejmrowiec anyone else wants to take a look?

PR #193, opened 2017-01-09
Make S3 region configurable via environment variable

GregorioDiStefano:
This is a useful in general, and I need it for testing :)

I want to be able to set the AWS region using an environment variable with docker-compose. The only other way would be to modify the config of the deployments config file.


maciejmrowiec:
All config options from config file can be set through environment.

GregorioDiStefano:
So DEPLOYMENTS_AWS.REGION should work?

GregorioDiStefano:
it does

PR #195, opened 2017-01-10
Fix pushing of "latest" image failing.

kacf:
It's now called "pr", so we need to retag it first.

Signed-off-by: Kristian Amlie <kristian.amlie@mender.io>


kacf:
@GregorioDiStefano: Once approved I will post the same for the other repositories.

GregorioDiStefano:
looks good @kacf

PR #197, opened 2017-01-10
relative url in Location header

mchalski:
Issues: MEN-881

see: https://tools.ietf.org/html/rfc3986#section-5.4

### DIFF
--- a/resources/images/view/view.go
+++ b/resources/images/view/view.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/ant0ine/go-json-rest/rest"
 	"github.com/mendersoftware/go-lib-micro/log"
@@ -38,7 +39,7 @@ type RESTView struct {
 }
 
 func (p *RESTView) RenderSuccessPost(w rest.ResponseWriter, r *rest.Request, id string) {
-	w.Header().Add(HttpHeaderLocation, fmt.Sprintf("%s/%s", r.URL.String(), id))
+	w.Header().Add(HttpHeaderLocation, fmt.Sprintf("./%s/%s", strings.TrimLeft(r.URL.Path, "/api/0.0.1/"), id))
 	w.WriteHeader(http.StatusCreated)
 }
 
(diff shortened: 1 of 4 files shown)


resources/deployments/controller/controller_deployments.go
```diff
@@ -65,7 +65,7 @@ func (d *DeploymentsController) PostDeployment(w rest.ResponseWriter, r *rest.Re
 		return
 	}
 
-	d.view.RenderSuccessPost(w, r, id)
+	d.view.RenderSuccessPost(w, r, "./deployments/"+id)
```
maciejmrowiec: adding header and it's content is self contained in view, why not modifying that instead?

resources/images/view/view.go
```diff
@@ -37,8 +36,8 @@ var (
 type RESTView struct {
 }
 
-func (p *RESTView) RenderSuccessPost(w rest.ResponseWriter, r *rest.Request, id string) {
-	w.Header().Add(HttpHeaderLocation, fmt.Sprintf("%s/%s", r.URL.String(), id))
```
maciejmrowiec: `r.URL.String() -> r.URL.Path`

would do the trick, if we would just align routing of gateway with service routing that should have been done :/

resources/deployments/controller/controller_deployments.go
maciejmrowiec: adding header and it's content is self contained in view, why not modifying that instead?

mchalski: well, the controller knows what resource is requested, not the view. the view is dumber than the controller, so it it needs to know what path exactly to set.

maciejmrowiec: My point was that controller also won't know anything specific about routing. This way we add yet another place that needs to be tweaked when messing around with route naming.

`w.Header().Add(HttpHeaderLocation, fmt.Sprintf("./%s/%s", strings.TrimLeft(r.URL.Path, "/api/0.0.1/), id))`

This way this will be universal and work for paths (outside of tweaks to common route).

When we align routing with gateway we can then simplify to `r.URL.Path/id`, no changes outside of module required.

What do you think ?

mchalski: got it, yes we can do that, as long as all Location-providing endpoints conform to the same pattern. they do currently, and no breaking changes are in sight, so I'll make a fix.

(I went for a more flexible solution after working on devadm - mixes base urls e.g. `/foo/{id}` and `foo/{id}/status`, so depending on the EP, the Location pattern is different).

mchalski: fixed

PR #198, opened 2017-01-11
Upload handling changed.

kjaskiewiczz:
Artifact uploaded to the deployments service will be uploaded to file
storage using presigned link and parsed by artifact library concurrently.
The solution is based on the io.TeeReader and io.Pipe.
Both operations needs to be done in the concurrent goroutines because
reading from the Pipe is a blocking operation.

@bboozzoo @maciejmrowiec @mchalski

### DIFF
--- a/resources/images/model/images.go
+++ b/resources/images/model/images.go
@@ -16,11 +16,16 @@ package model
 
 import (
 	"io"
+	"io/ioutil"
 	"time"
 
 	"github.com/mendersoftware/deployments/resources/images"
 	"github.com/mendersoftware/deployments/resources/images/controller"
+	"github.com/mendersoftware/mender-artifact/metadata"
+	"github.com/mendersoftware/mender-artifact/parser"
+	"github.com/mendersoftware/mender-artifact/reader"
 	"github.com/pkg/errors"
+	"github.com/satori/go.uuid"
 )
 
 const (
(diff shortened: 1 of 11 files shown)


resources/images/model/images.go
```diff
@@ -45,42 +50,102 @@ func NewImagesModel(
...
+	pR, pW := io.Pipe()
+	// limit reader to max image size
+	lr := io.LimitReader(imageReader, MaxImageSize)
+	tee := io.TeeReader(lr, pW)
+
+	artifactFileID := uuid.NewV4().String()
+
+	ch := make(chan error)
+	// create goroutine for artifact upload
+	go func() {
```
bboozzoo: IMO there should be a comment explaining why this is done concurrently inside goroutine

resources/images/model/images.go
```diff
@@ -45,42 +50,102 @@ func NewImagesModel(
...
+		return "", "", err
+	}
+
+	// close the pipe
+	pW.Close()
 
-	if err := metaConstructor.Validate(); err != nil {
-		return "", controller.ErrModelInvalidMetadata
+	// collect output from the goroutine
+	uploadResponseErr := <-ch
```
bboozzoo:
```
if uploadResponseErr := <- ch; uploadResponseErr != nil {
    return "", "", uploadResponseErr
}
```

resources/images/model/images.go
```diff
@@ -45,42 +50,102 @@ func NewImagesModel(
...
+	if metaConstructor == nil {
 		return "", controller.ErrModelMissingInputMetadata
 	}
+	if imageReader == nil {
+		return "", controller.ErrModelMissingInputArtifact
+	}
+	imageID, artifactFileID, err := i.handleArtifact(metaConstructor, imageReader)
+	// try to remove artifact file from file storage on error
+	if err != nil {
+		i.fileStorage.Delete(artifactFileID)
```
maciejmrowiec: In case of failed cleanup it would be nice to be able to catch this in logs, maybe wrapping in error from Delete into `err` would be an option.

resources/images/model/images.go
```diff
@@ -45,42 +50,102 @@ func NewImagesModel(
...
+	if err != nil {
+		pW.Close()
+		_ = <-ch
+		return "", "", controller.ErrModelInvalidMetadata
+	}
+
+	// read the rest of the data, just for safety
+	_, err = io.Copy(ioutil.Discard, tee)
+	if err != nil {
+		pW.Close()
```
maciejmrowiec: `defer pW.Close()` after declaring pipe may save you few lines

resources/images/model/images.go
```diff
@@ -45,42 +50,102 @@ func NewImagesModel(
...
+	go func() {
+		err := i.fileStorage.UploadArtifact(artifactFileID, pR, ImageContentType)
+		ch <- err
+	}()
+
+	// parse artifact
+	metaArtifactConstructor, err := getMetaFromArchive(&tee, MaxImageSize)
+	if err != nil {
+		pW.Close()
+		_ = <-ch
```
maciejmrowiec: alternatively https://golang.org/pkg/sync/#WaitGroup if in need to wait for the job to finish

resources/images/model/images.go
kjaskiewiczz: done

resources/images/model/images.go
kjaskiewiczz: done

resources/images/model/images.go
kjaskiewiczz: done

resources/images/model/images.go
kjaskiewiczz: I'm not sure if there will be a profit from such solution, since we are closing PipeWriter in the middle of this method.

resources/images/model/images.go
kjaskiewiczz: the error handling here was extended

bboozzoo:
Looks like the image was uploadded here: https://travis-ci.org/mendersoftware/deployments/jobs/191244190#L2655 but when the device tries to download it using the link it got from deployments, we get back 404 https://travis-ci.org/mendersoftware/deployments/jobs/191244190#L2701

resources/images/model/images.go
maciejmrowiec: main benefit is that you don't have to remember to do it later in the code (less code, harder to make a bug while tweaking this later), this is more of convenience thing (only suggestion)

resources/images/model/images.go
```diff
@@ -45,42 +50,115 @@ func NewImagesModel(
...
+	// max image size - 10G
+	const MaxImageSize = 1024 * 1024 * 1024 * 10
+
+	// create pipe
+	pR, pW := io.Pipe()
+	// limit reader to max image size
+	lr := io.LimitReader(imageReader, MaxImageSize)
+	tee := io.TeeReader(lr, pW)
+
+	artifactFileID := uuid.NewV4().String()
```
maciejmrowiec: Using separate ID for storage and fileserver is an issue, it artifact have to be signified with the same ID across storages otherwise you won't be able to access this file.

resources/images/model/images.go
```diff
@@ -45,42 +50,115 @@ func NewImagesModel(
...
+
+	// validate artifact metadata
+	if err = metaArtifactConstructor.Validate(); err != nil {
+		return "", "", controller.ErrModelInvalidMetadata
 	}
+
+	// check if artifact is unique
+	// artifact is considered to be unique if there is no artifact with the same name
+	// and supporing the same platform in the system
 	isArtifactUnique, err := i.imagesStorage.IsArtifactUnique(
```
maciejmrowiec: Would be nice if we would cancel upload to fileserver if we figure out from meta this is not going to fly. http.Client has some build in cancel functionality through the `context` package. But this is convenience thing. Not sure how much coding this would take, we could do this now if you think so, or leave a TODO in code that it's possible to optimise this in the future.

kjaskiewiczz: Currently, while parsing artifact, artifact library is reading whole artifact file. It would be nice to know if we should upload artifact to the file storage after just reading artifact metadata, but this will need some support in the artifact library IMO.

resources/images/model/images.go
kjaskiewiczz: I'm not sure what is the overhead of closing closed writer.

resources/images/model/images.go
maciejmrowiec: You can read stream section by section, pull name&devicetype from headers and do early validation and send cancel to upload routine, currently you need to upload entire 2GB file just to figure out that it have been already uploaded. 

Example test from artifact lib: https://github.com/mendersoftware/mender-artifact/blob/master/reader/reader_test.go#L233

kjaskiewiczz: I've been talking exactly about this on the backend team meeting yesterday.
It's possible to do it like that, but it will be better to have support for things like that in the artifact library (e.g. callback called when this data was read). Currently it will looks more like hacking this library then using it in a proper way IMO.
I think there should be another task for this (not bug but improvement).

kjaskiewiczz: We can improve artifact library and move all required data to the metadata, maybe?

maciejmrowiec: Yes, we can do this as another patch, it's not related that much to the issue we have at hand now.

resources/images/model/images.go
kjaskiewiczz: fixed

PR #203, opened 2017-01-19
Configuration clean & vendoring fixes

bboozzoo:
Cleanup of configuration options. Also had to update `github.com/spf13/viper` as a result to include 2 fixes that improve environment binds.

Also includes some pre-spring cleaning of `/vendor`. It turns out that we had a large number of deps that were not used anymore and dependencies that were not included (but should have been). A commit that skips `install` step in travis will help to reveal what's still missing from `/vendor`.


main.go
```diff
@@ -54,13 +55,17 @@ func HandleConfigFile(filePath string) (config.ConfigReader, error) {
...
+	c.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
+
 	//Allow AWS URI endpoint to be set by environment variable
 	c.BindEnv(SettingAwsURI, "AWS_URI")
 
 	//Allow aws keyid, aws token, aws secret to be read by viper
 	c.BindEnv(SettingAwsAuthKeyId, "AWS_ACCESS_KEY_ID")
 	c.BindEnv(SettingAwsAuthSecret, "AWS_SECRET_ACCESS_KEY")
 	c.BindEnv(SettingAwsAuthToken, "AWS_SESSION_TOKEN")
+	c.BindEnv(SettingAwsS3Region, "AWS_S3_REGION")
```
maciejmrowiec: with MEN-747 all of those supposed to get thrown away, not sure if there is a point introducing this now

bboozzoo: I'm not sure I fully understood both tickets, but in [MEN-747](https://tracker.mender.io/browse/MEN-747?focusedCommentId=77578&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-77578) you mentioned:
> ah, sorry if this was not that clear, was assuming we will still maintain functionality of overriding config settings with environment variables as we are able to do today. As you said this is very convenient, especially in out current setup. 

Removing these options will not preserve config through environment functionality.

maciejmrowiec: There is a global binding of env variables co config options that this PR is also fixing few lines above. You should be able to set any config setting with `export DEPLOYMENTS_<option path>=<value>`

What we don't need is two different variables that set the same thing and people wasting time trying to figure out how to use them and how they differ.

bboozzoo: What's the advantage of renaming already known & used AWS_* keys to DEPLOYMENTS_AWS_* and updating documentation, READMEs, integration setup configuratoin and possibly integration tests?

Personally, I find those 4 AWS_* environment keys quite clear, but if you feel we should rename those to DEPLOYMENTS_AWS_ and update all places that refer to this then ok. It's just that this does not look like a 3SP job to me.

bboozzoo: ping anyone? this is kind of blocking both tasks

maciejmrowiec: What I'm after is a singe, unified, consistent way to deal with settings. Having a simple doc mention: `if you want to set any available configuration option though ENV just set env var with servciename_settingname variable, all CAPS, while accessing nested settings use _ or / whatever to indicate nesting, for available options look at config.yaml for specific service` and we just documented clearly how this use it.

bboozzoo:
@GregorioDiStefano heads up, environment keys change, this may break some tests

PR #204, opened 2017-01-19
introduce and use the middleware setting (defaults to prod)

mchalski:
by default, `deployments` would spit out stack traces on panics - this was because the dev middleware stack was used by default (with `RecoverMiddleware` having `EnableResponseStackTrace: true`).

the dev/prod distinction was missing, so this PR introduces it (with `prod` as the default, i.e. no stack traces, just a generic "internal server error").

Issues: MEN-898

### DIFF
--- a/config.go
+++ b/config.go
@@ -45,6 +45,9 @@ const (
 
 	SettingGateway        = "mender-gateway"
 	SettingGatewayDefault = "localhost:9080"
+
+	SettingMiddleware        = "middleware"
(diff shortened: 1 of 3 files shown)


bboozzoo:
The PR is already approve but, for some unknown reason, is waiting for codecov to post status report (one was already posted 22hrs ago). I don't think that a new codecov status report is ever going to come. Can someone merge this?

PR #205, opened 2017-01-19
Handle "size" part of the mulipart/form-data form from the artifact upload request.

kjaskiewiczz:
Support for the new part ("size") of the multipart/form-data form from artifact upload request.
Fixes MEN-957

TODO:
more tests needed!

@maciejmrowiec @bboozzoo @mchalski

### DIFF
--- a/docs/management_api.yml
+++ b/docs/management_api.yml
@@ -308,6 +308,11 @@ paths:
       consumes:
         - multipart/form-data
       parameters:
+        - name: size
+          in: formData
+          description: Size of the artifact file in megabytes.
+          required: true
+          type: long
         - name: description
           in: formData
           required: false
--- a/resources/images/controller/images.go
+++ b/resources/images/controller/images.go
@@ -272,7 +272,9 @@ func (s *SoftwareImagesController) NewImage(w rest.ResponseWriter, r *rest.Reque
 	case ErrModelArtifactNotUnique:
 		l.Error(err.Error())
 		s.view.RenderError(w, r, cause, http.StatusUnprocessableEntity, l)
-	case ErrModelMissingInputMetadata, ErrModelMissingInputArtifact, ErrModelInvalidMetadata:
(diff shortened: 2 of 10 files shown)


docs/management_api.yml
```diff
@@ -308,6 +308,11 @@ paths:
       consumes:
         - multipart/form-data
       parameters:
+        - name: size
+          in: formData
+          description: Size of the artifact file.
```
bboozzoo: in bytes right?

bboozzoo: BTW. i think it should be `long` instead of integer. The code expects `int64`and according to swagger spec http://swagger.io/specification/#data-types-12 integer is int32

resources/images/controller/images.go
```diff
@@ -245,54 +256,68 @@ func (s *SoftwareImagesController) NewImage(w rest.ResponseWriter, r *rest.Reque
...
+			// valide metadata provided by the user and the image size
+			if err := multipartUploadMsg.MetaConstructor.Validate(); err != nil {
+				return nil, err
+			}
+			// artifact size part should be provided before artifact part
 			// HTML form can't set specific content-type, it's automatic, if not empty - it's a file
-			if p.Header.Get("Content-Type") == "" {
-				return nil, nil, errors.New("The last part of the multipart/form-data message should be an image.")
+			if multipartUploadMsg.ArtifactSize == 0 || p.Header.Get("Content-Type") == "" {
+				return nil, errors.New("The last part of the multipart/form-data message should be an artifact.")
```
bboozzoo: the error message is not related to `multipartUploadMsg.ArtifactSize == 0` check

bboozzoo: BTW. a negative size will validate ok?

docs/management_api.yml
kjaskiewiczz: fixed

resources/images/controller/images.go
kjaskiewiczz: fixed

PR #206, opened 2017-01-20
fix Deployment.GetStatus() 

mchalski:
Issue: https://tracker.mender.io/browse/MEN-935

the crux of the issue is that the `already-installed` state is not included in our checks when calling `Get /deployments/:id` - see jira ticket for discussion.

solution: refactor and simplify the GetStatus() method to correctly (and more explicitly) handle all state combinations.

### DIFF
--- a/resources/deployments/deployment.go
+++ b/resources/deployments/deployment.go
@@ -138,7 +138,13 @@ func (d *Deployment) IsInProgress() bool {
 		acount += d.Stats[s]
 	}
 
-	if acount != 0 {
+	if acount != 0 ||
+		(d.Stats[DeviceDeploymentStatusPending] > 0 &&
+			(d.Stats[DeviceDeploymentStatusAlreadyInst] > 0 ||
+				d.Stats[DeviceDeploymentStatusSuccess] > 0 ||
+				d.Stats[DeviceDeploymentStatusFailure] > 0 ||
+				d.Stats[DeviceDeploymentStatusNoArtifact] > 0 ||
+				d.Stats[DeviceDeploymentStatusAborted] > 0)) {
 		return true
 	}
 	return false
@@ -168,26 +174,31 @@ func (d *Deployment) IsFinished() bool {
 }
 
 func (d *Deployment) IsPending() bool {
-	if d.IsInProgress() {
-		return false
+	//pending > 0, evt else == 0
+	if d.Stats[DeviceDeploymentStatusPending] > 0 &&
+		d.Stats[DeviceDeploymentStatusDownloading] == 0 &&
+		d.Stats[DeviceDeploymentStatusInstalling] == 0 &&
+		d.Stats[DeviceDeploymentStatusRebooting] == 0 &&
+		d.Stats[DeviceDeploymentStatusSuccess] == 0 &&
+		d.Stats[DeviceDeploymentStatusAlreadyInst] == 0 &&
+		d.Stats[DeviceDeploymentStatusFailure] == 0 &&
+		d.Stats[DeviceDeploymentStatusNoArtifact] == 0 {
+
+		return true
 	}
 
-	if d.IsFinished() {
-		return false
-	}
-
-	return true
+	return false
 }
 
 func (d *Deployment) GetStatus() string {
 	if d.IsAborted() {
 		return "aborted"
-	} else if d.IsInProgress() {
-		return "inprogress"
+	} else if d.IsPending() {
+		return "pending"
 	} else if d.IsFinished() {
 		return "finished"
 	} else {
-		return "pending"
+		return "inprogress"
 	}
 }
 
--- a/resources/deployments/deployment.go
+++ b/resources/deployments/deployment.go
@@ -188,9 +188,7 @@ func (d *Deployment) IsPending() bool {
 }
 
 func (d *Deployment) GetStatus() string {
-	if d.IsAborted() {
-		return "aborted"
-	} else if d.IsPending() {
+	if d.IsPending() {
 		return "pending"
 	} else if d.IsFinished() {
 		return "finished"
--- a/resources/deployments/deployment.go
+++ b/resources/deployments/deployment.go
@@ -160,17 +160,14 @@ func (d *Deployment) IsAborted() bool {
 }
 
 func (d *Deployment) IsFinished() bool {
-	// check if there are downloading/rebooting/installing devices
-	if d.IsInProgress() {
-		return false
+	if d.Stats[DeviceDeploymentStatusPending] == 0 &&
+		d.Stats[DeviceDeploymentStatusDownloading] == 0 &&
+		d.Stats[DeviceDeploymentStatusInstalling] == 0 &&
+		d.Stats[DeviceDeploymentStatusRebooting] == 0 {
+		return true
 	}
 
-	// check if there are pending devices
-	if d.Stats[DeviceDeploymentStatusPending] != 0 {
-		return false
-	}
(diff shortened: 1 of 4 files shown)


resources/deployments/deployment.go
```diff
@@ -123,6 +124,8 @@ func (d *Deployment) MarshalJSON() ([]byte, error) {
 		Status:  d.GetStatus(),
 	}
 
+	fmt.Printf("marshaller got status: %v for deployment: %#+v\n", slim.Status, d)
```
bboozzoo: clean that up please

resources/deployments/deployment.go
```diff
@@ -164,30 +167,36 @@ func (d *Deployment) IsFinished() bool {
 		return false
 	}
 
+	fmt.Printf("IsFinished returning true for stats %v\n", d.Stats)
```
bboozzoo: this one too

maciejmrowiec:
looks good, outside of prints pointed out by @bboozzoo 

btw will this need any adjustments to: https://github.com/mendersoftware/deployments/blob/master/resources/deployments/mongo/deployments.go#L241 ?

resources/deployments/deployment.go
mchalski: done

resources/deployments/deployment.go
mchalski: done

mchalski:
@maciejmrowiec imo this part is ok

maciejmrowiec:
@mchalski 

When requesting for status=inprogress, and having case: pending=1, already-instralled=1, success=1, everything else 0, this query won't match any results if I'm not mistaken
```
case deployments.StatusQueryInProgress:
		{
			// downloading, installing or rebooting are non 0
			stq = bson.M{
				"$or": []bson.M{
					bson.M{
						buildStatusKey(deployments.DeviceDeploymentStatusDownloading): gt0,
					},
					bson.M{
						buildStatusKey(deployments.DeviceDeploymentStatusInstalling): gt0,
					},
					bson.M{
						buildStatusKey(deployments.DeviceDeploymentStatusRebooting): gt0,
					},
				},
			}
		}
```

mchalski:
@maciejmrowiec right. hold on, I'll do another rework

resources/deployments/mongo/deployments.go
```diff
@@ -261,6 +261,9 @@ func buildStatusQuery(status deployments.StatusQuery) bson.M {
 					bson.M{
 						buildStatusKey(deployments.DeviceDeploymentStatusRebooting): gt0,
 					},
+					bson.M{
+						buildStatusKey(deployments.DeviceDeploymentStatusAlreadyInst): gt0,
```
maciejmrowiec: not sure if this will do the trick, consider object  with:

```
already-installed:1
everything else: 0
```

This would end up in results of `inprogress`  while should be in finished.

```
pending: 1
success: 1
```

This should end up in `inprogress` results while it won't be included.

mchalski: ugh, that's correct. should be fixed now.

resources/deployments/mongo/deployments.go
```diff
@@ -261,9 +262,20 @@ func buildStatusQuery(status deployments.StatusQuery) bson.M {
 					bson.M{
 						buildStatusKey(deployments.DeviceDeploymentStatusRebooting): gt0,
 					},
+					bson.M{
+						"$and": []bson.M{
+							bson.M{
+								buildStatusKey(deployments.DeviceDeploymentStatusAlreadyInst): gt0,
```
maciejmrowiec: that needs to cover already installed, success, failure, noimage (all of them are finished with with pending > 0 are in progress)

dammit those cases are growing, btw that function is covered with unit test, maybe it's a good option to add cases there

maciejmrowiec:
Lets write down cases we have:

pending:

```
pending > 0 AND
downloading = 0 AND
installing = 0 AND
rebooting = 0 AND
already-installed = 0 AND
noimage = 0 AND
failure = 0 AND
success = 0
```

finished
```
pending = 0 AND
downloading = 0 AND
installing = 0 AND
rebooting = 0 AND (
already-installed > 0 OR
noimage > 0 OR
failure > 0 OR
success > 0 )
```

inprogress:

```
(downloading > 0 OR installing > 0 OR rebooting > 0) OR (pending > 0 AND (already-installed > 0 OR noimage > 0 OR failure > 0 OR success > 0))
```

does that make sense, or am I missing some?

resources/deployments/mongo/deployments.go
mchalski: yea it sucks. ok, I can add the two cases you mentioned.

mchalski:
@maciejmrowiec we've finally come up with an exact truth table to chase these states down:
* cols - device deployment state (pending, downloading, installing, rebooting, success, failure, noartifact, already-installed, aborted), 
* rows - deployment state (pending, finished, inprogress, aborted)
* `1` means `counter greater than 0`
* `x` means 'don't care'

|    | P | D | I | R | S | F | NA | AI | A |
|----|---|---|---|---|---|---|----|----|---|
| P  | 1 | 0 | 0 | 0 | 0 | 0 | x  | 0  | 0 |
| F  | 0 | 0 | 0 | 0 | x | x | x  | x  | x |
| IP | x | 1 | 1 | 1 | x | x | x  | x  | x |
| A  | x | x | x | x | x | x | x  | x  | 1 |

and to answer your question:
* one missing state is `aborted` (if any dev has deployment 'aborted')
* `pending` is correct
* `finished` can be reformulated as just: pending=0 and downloading=0 and installing=0 and rebooting=0 (all other states count towards `finished`)
* `inprogress` is correct


ok, I'll the fix for `inprogress` and some tests

maciejmrowiec:
aborted counts towards finished

maciejmrowiec:
that table skips `pending+finished` cases for inprogress

mchalski:
@maciejmrowiec 

re table
- hmm, yes, disregard

re aborted state
- it does count as 'finished' when supplied in the query filter (filter doesn't accept an 'aborted' state) but
- when asking for a single deployment - it's `status` will be `aborted` if >0 devs have the `aborted` status - that's what I was referring to

so there is one status that we can get out of `GET /deployments/:id` but not `GET /deployments?query=....`

maciejmrowiec:
Both: `GET /deployments/:id` and `GET /deployments?query...` support only `pending`, `inprogress` and `finished` https://docs.mender.io/APIs/Integration-APIs/Deployments#deployment

BTW I see implementation mismatch documentation (code will return aborted while we are documenting only 3 statuses). Question is which one is the desired behaviour? - GUI does not need to set apart finished from aborted on that high level (only in details when it accesses more detailed stats), so I would lean towards taking the documented one.

mchalski:
@maciejmrowiec yes, there's a mismatch, I have no strong opinion - @michaelatmender should say if he expects and handles the `aborted` state currently

mchalski:
@maciejmrowiec ok, here's a last batch of updates and corrections:
- added some tests for cases mentioned in this conversation; 
- noticed `GET /deployments?query=...` has an unnecesarily convoluted condition for the `finished` filter, refactored
- made `Deployment.IsFinished()` match the above exactly, so that we don;t have to chase diffs in these two pieces of code
- removed the `aborted` deployment state, now it counts towards `finished`; the docs stay the same

is there anything else?

maciejmrowiec:
Looks good, thanks for resolving this once for all!

PR #209, opened 2017-02-01
MEN-963: Update API versioning.

maciejmrowiec:
Change public routing to v1, v2, v3 schema (v1).
Update documentation.
Remove port 8080 from swagger.
Add service version header returned by API

@mendersoftware/rndity

### DIFF
--- a/docs/devices_api.yml
+++ b/docs/devices_api.yml
@@ -1,14 +1,14 @@
 swagger: '2.0'
 info:
   title: Deployments
-  version: '0.1'
+  version: '1'
   description: |
     An API for device firmware deployments. Intended for use by devices.
 
     Devices can get new updates and send information about current deployment status.
 
-host: 'docker.mender.io:8080'
-basePath: '/api/devices/0.1/deployments'
+host: 'docker.mender.io'
+basePath: '/api/devices/v1/deployments'
 schemes:
   - https
 
--- a/middleware.go
+++ b/middleware.go
@@ -21,6 +21,7 @@ import (
 	"github.com/ant0ine/go-json-rest/rest"
 	"github.com/mendersoftware/deployments/config"
 	"github.com/mendersoftware/go-lib-micro/accesslog"
+	"github.com/mendersoftware/go-lib-micro/customheader"
 	"github.com/mendersoftware/go-lib-micro/requestid"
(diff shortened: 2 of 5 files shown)


middleware.go
```diff
@@ -77,6 +78,12 @@ var DefaultProdStack = []rest.Middleware{
 }
 
 func SetupMiddleware(c config.ConfigReader, api *rest.Api) {
+
+	api.Use(&customheader.CustomHeaderMiddleware{
+		HeaderName:  "X-DEPLOYMENTS-VERSION",
+		HeaderValue: CreateVersionString(),
```
bboozzoo: If you build locally this becomes `X-DEPLOYMENTS-VERSION: Unknown`

docs/devices_api.yml
```diff
@@ -1,14 +1,14 @@
...
-  version: '0.1'
+  version: '1'
   description: |
     An API for device firmware deployments. Intended for use by devices.
 
     Devices can get new updates and send information about current deployment status.
 
-host: 'docker.mender.io:8080'
-basePath: '/api/devices/0.1/deployments'
+host: 'docker.mender.io'
```
bboozzoo: BTW. maybe we should not provide the host at all, since it makes little sense. If I start the service locally, there's no way I will access it by going to http://docker.mender.io.

middleware.go
maciejmrowiec: Yes thats when variables are not set during build, on locally this can be done if someone needs to, but otherwise "Unknown" to make this clear.

docs/devices_api.yml
maciejmrowiec: This is just for example, when we have hosted version I would definitely change this real domain we will use to make it copy-paste.

middleware.go
bboozzoo: Ok, but in regular builds it will also look a bit awkward. For instance master builds will return `master_a3e9adbf54fb4f606377ef134f211fe65a6224ab`. If there's a tag we'll get `0.0.2` or `integration-0.1` (these are currently defined in deployments repo). I was under the assumption that we want to have a service API level rather than service version.

maciejmrowiec: thats application version

maciejmrowiec:
merging for now to get things going, but let's continue the thread we there is a need

PR #217, opened 2017-02-08
MEN-1003: Multiple status submissions causing issues 

maciejmrowiec:
@mchalski

### DIFF
--- a/resources/deployments/deployment_external_test.go
+++ b/resources/deployments/deployment_external_test.go
@@ -15,6 +15,7 @@
 package deployments_test
 
 import (
+	"math/rand"
 	"testing"
 	"time"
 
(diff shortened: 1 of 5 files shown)


resources/deployments/deployment_external_test.go
```diff
@@ -360,3 +361,45 @@ func TestDeploymentGetStatus(t *testing.T) {
...
+
+		if dep.GetStatus() == "finished" {
+			finished++
+		}
+
+		if dep.GetStatus() == "inprogress" {
+			inprogress++
+		}
+
+		exp_stats := pending + inprogress + finished
```
pasinskim: Minor: I guess we don't need a separate variable here. You can `assert.Equal(t, 1, pending + inprogress + finished, dep.Stats)`.

resources/deployments/mongo/deployments_external_test.go
```diff
@@ -549,6 +549,60 @@ func TestDeploymentStorageUpdateStats(t *testing.T) {
 			OutputError: ErrStorageInvalidID,
 			OutputStats: nil,
 		},
+		"no old state": {
+			InputID: "a108ae14-bb4e-455f-9b40-2ef4bab97bb7",
+			InputDeployment: &deployments.Deployment{
+				Id: StringToPointer("a108ae14-bb4e-455f-9b40-2ef4bab97bb7"),
+				Stats: map[string]int{
```
pasinskim: This map is repeated 3 times here. Maybe we can have `s := map ...` and use this instead.

resources/deployments/deployment_external_test.go
maciejmrowiec: line was to long ;)

pasinskim: It will be optimized anyway by the compiler I guess...

resources/deployments/mongo/deployments_external_test.go
maciejmrowiec: I prefer to keep test cases independent of each other.

PR #226, opened 2017-02-15
Fix missing branch problem on non-master branches.

kacf:
Signed-off-by: Kristian Amlie <kristian.amlie@mender.io>


kacf:
Change has no effect on integration tests, and they failed because #224 is missing, but Travis part passed.

PR #227, opened 2017-02-15
Fix missing branch problem on non-master branches.

kacf:


kacf:
This has no effect on integration tests, so cancelled that part. The Travis part passed.

PR #229, opened 2017-02-15
Update 1.0.x

pasinskim:


maciejmrowiec:
some of those are already in that branch (there ware cherry-picks)

pasinskim:
All important changes seems to be cherry-picked.

PR #243, opened 2017-03-01
API docs: add pagination to GET deployments endpoint

kjaskiewiczz:


maciejmrowiec:
Do we miss link header mention ?

kjaskiewiczz:
Link header added

PR #250, opened 2017-03-28
Update artifact handling reflecting changes in mender-artifact.

pasinskim:
Changelog: Title

Signed-off-by: Marcin Pasinski <marcin.pasinski@mender.io>


kjaskiewiczz:
https://github.com/mendersoftware/deployments/blob/master/docs/management_api.yml#L768
Should we update API spec?

pasinskim:
@kjaskiewiczz at some point we should. There is a separate task for the backend: https://tracker.mender.io/browse/MEN-1022. We can fix this as a part of that task. This one was supposed to reflect latest changes in mender-artifact library.

PR #262, opened 2017-05-05
Increase file upload request validity when pushing artifat to remote …

maciejmrowiec:
…file storage.

Change presign valitiry from 10 seconds to 1 minute.

Changelog: Title

### DIFF
--- a/resources/images/s3/filestorage.go
+++ b/resources/images/s3/filestorage.go
@@ -180,7 +180,7 @@ func (s *SimpleStorageService) UploadArtifact(ctx context.Context,
 	r, _ := s.client.PutObjectRequest(params)
 
 	// Presign request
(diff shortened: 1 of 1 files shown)


maciejmrowiec:
Fixed type, updated commit message, changed to 5 minutes.

This proposal is due to suspicion that one of the users hit time synchronisation issue between fileserver and deployments service which results in issuing presigned links that are expired straight away. Making them last for few minutes should mitigate smaller de-synchronizations.

kjaskiewiczz:
https://groups.google.com/a/lists.mender.io/forum/#!msg/mender/UeKbra-XQ3Y/YM1tE6ZICgAJ
looks like he's using minio...

bboozzoo:
I suppose we can merge it and see what happens. We'll dig deeper If problems persist.

PR #267, opened 2017-05-23
New deployment creation process

kjaskiewiczz:
Description can be found here: https://tracker.mender.io/browse/MEN-1185


resources/deployments/controller/controller_deployments.go
```diff
@@ -63,7 +64,11 @@ func (d *DeploymentsController) PostDeployment(w rest.ResponseWriter, r *rest.Re
 
 	id, err := d.model.CreateDeployment(ctx, constructor)
 	if err != nil {
-		d.view.RenderInternalError(w, r, err, l)
+		if err == ErrNoArtifact {
+			d.view.RenderError(w, r, err, http.StatusBadRequest, l)
```
bboozzoo: How about 422 Unprocessable Entity instead?

resources/deployments/model/deployments.go
```diff
@@ -194,18 +302,26 @@ func (d *DeploymentsModel) GetDeploymentForDeviceWithCurrent(ctx context.Context
 		return nil, nil
 	}
 
-	link, err := d.imageLinker.GetRequest(ctx, deployment.Image.Id,
+	if err := d.assignArtifact(ctx, deployment, deviceDeployment, installed); err != nil {
```
bboozzoo: I'm not sure if this is ok to assign a new image each time we reach this place. I'd guess that once an image is assigned you could skip looking up a new image. For instance suppose that devices are retrying image download by first asking to see if there is a deployment. This would make sense if device type can change in between the requests. @estenberg @maciejmrowiec is this possible?

maciejmrowiec: We don't forbid it on API so it could happen, mainly depends on client implementation. Doe in the example you are mentioning I would not imagine why it would change.

resources/deployments/model/deployments.go
```diff
@@ -156,19 +189,85 @@ func (d *DeploymentsModel) ImageUsedInActiveDeployment(ctx context.Context,
...
+	// If not having appropriate image, set noartifact status
+	if artifact == nil {
+		if err := d.UpdateDeviceDeploymentStatus(ctx, *deviceDeployment.DeploymentId,
+			*deviceDeployment.DeviceId, deployments.DeviceDeploymentStatusNoArtifact); err != nil {
+			return errors.Wrap(err, "Failed to update deployment status")
+		}
+		return nil
+	}
+
+	if err := d.deviceDeploymentsStorage.AssignArtifact(
```
maciejmrowiec: if we have already assigned image (that we reset to `nil` in line 224), and we and up assigning the same one, could we skip this update?

kjaskiewiczz: if the device type is the same as reported previously and the artifact is already assigned we can skip reassignment of the artifact

resources/deployments/controller/controller_deployments.go
kjaskiewiczz: done

resources/deployments/model/deployments.go
kjaskiewiczz: done

kjaskiewiczz:
Integration tests are failing.
@GregorioDiStefano can you update integration tests?

PR #268, opened 2017-06-01
Add basic support for artifact containing state scripts.

pasinskim:


resources/images/model/images.go
```diff
@@ -341,6 +342,11 @@ func getMetaFromArchive(r *io.Reader) (*images.SoftwareImageMetaArtifactConstruc
 		return nil
 	}
 
+        aReader.ScriptsReadCallback = func(r io.Reader, fi os.FileInfo) error {
```
maciejmrowiec: Do we need this callback? - looks like piece of unused code that may stick around for a long time.

vendor/vendor.json
```diff
@@ -441,28 +441,28 @@
 			"revisionTime": "2017-04-20T09:04:30Z"
 		},
 		{
-			"checksumSHA1": "vrp0KSG3kZimXujiJMc/NTf2w3M=",
+			"checksumSHA1": "+Yj9sWqspkwTTYsSpnDACT7ZPU0=",
```
maciejmrowiec: maybe we should start vendor artefact lib from artefact lib release tags?
now as I look at it I'm not sure which artifact spec version we support after this merge

could be nice to mention this also in api doc (upload supports artifact spec xyz, it's compatible with 1,2,3), looks also like this is a new functionality which may need changelog entry (maybe?)

resources/images/model/images.go
pasinskim: We do not, but this was to simplify the integration in future. There are some tasks for the backend and maybe we will show scripts in UI.

vendor/vendor.json
pasinskim: That's a good idea, but this is how it was for now. What is more in master this might be tricky unless we will be able to have master here and always take latest artifact library changes.

maciejmrowiec: Why master it would be trickier? 3d party dependency gets new release and once we decide to update to it we do it first in master.

resources/images/model/images.go
maciejmrowiec: No active plans to do this for now

pasinskim: Removed.

pasinskim:
@maciejmrowiec all right, the unnecessary commit was removed.

PR #279, opened 2017-08-25
Include Mongo SSL support in database migration code.

maciejmrowiec:
missing ssl setup from migration code


maciejmrowiec:
@GregorioDiStefano looks like that space removal done the trick, waiting for integration tests and merging

PR #280, opened 2017-09-04
multitenant migrations, server --automigrate command 

kjaskiewiczz:
https://tracker.mender.io/browse/MEN-1252

requires: https://github.com/mendersoftware/integration/pull/343

### DIFF
--- a/main.go
+++ b/main.go
@@ -45,6 +45,21 @@ func doMain(args []string) {
 		},
 	}
 
+	app.Commands = []cli.Command{
+		{
+			Name:  "server",
+			Usage: "Run the service as a server",
+			Flags: []cli.Flag{
+				cli.BoolFlag{
+					Name:  "automigrate",
+					Usage: "Run database migrations before starting.",
+				},
+			},
+
+			Action: cmdServer,
+		},
+	}
+
 	app.Action = cmdServer
 	app.Before = func(args *cli.Context) error {
 
@@ -79,6 +94,15 @@ func cmdServer(args *cli.Context) error {
 		return cli.NewExitError(
 			fmt.Sprintf("failed to connect to db: %v", err),
 			3)
+	}
+
+	if args.Bool("automigrate") {
+		err = MigrateDb(context.Background(), DbVersion, nil, dbSession)
+		if err != nil {
+			return cli.NewExitError(
+				fmt.Sprintf("failed to run migrations: %v", err),
+				3)
+		}
 	}
 
 	l.Printf("Deployments Service, version %s starting up",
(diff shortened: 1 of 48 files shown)


main.go
```diff
@@ -16,91 +16,111 @@ package main
...
-		l.Fatalf("failed to migrate DB data: %v", err)
+		err := config.FromConfigFile(configPath, configDefaults)
+		if err != nil {
+			return cli.NewExitError(
+				fmt.Sprintf("error loading configuration: %s", err),
+				1)
+		}
+
+		// Enable setting config values by environment variables
+		config.Config.SetEnvPrefix("DEPLOYMENTS")
```
bboozzoo: `config.FromConfigFile()` already does that

main.go
```diff
@@ -16,91 +16,111 @@ package main
...
-	SetDefaultConfigs(c)
+	l.Printf("Deployments Service, version %s starting up",
+		CreateVersionString())
 
-	// Find and read the config file
-	if filePath != "" {
-		c.SetConfigFile(filePath)
-		if err := c.ReadInConfig(); err != nil {
-			return nil, err
+	dbSession, err := mgo.Dial(config.Config.GetString(SettingMongo))
```
bboozzoo: It might also be the case in other services, but shouldn't we do `dbSession.Close()` once we're done with migration?

main.go
kjaskiewiczz: fixed

main.go
kjaskiewiczz: done
In other services migrations are using main datastore session (not a copy).

bboozzoo: ok

main.go
```diff
@@ -16,91 +16,108 @@ package main
...
```
maciejmrowiec: There is a helper function for opening connection with ssl support, use `NewMongoSession` instead

main.go
```diff
@@ -16,91 +16,108 @@ package main
...
+	if err != nil {
+		return cli.NewExitError(
+			fmt.Sprintf("failed to connect to db: %v", err),
+			3)
+	}
+	dbSession.SetSafe(&mgo.Safe{
```
maciejmrowiec: This is handled by NewMongoSession

main.go
kjaskiewiczz: fixed
I've been rebasing on top of this change (dial + set safe -> NewMongoSession) and resolve conflicts incorrectly.

main.go
kjaskiewiczz: fixed

bboozzoo:
merge?

PR #282, opened 2017-09-14
Fix index len limit

mchalski:
Issue: https://tracker.mender.io/browse/MEN-1415


GregorioDiStefano:
@maciejmrowiec @mchalski deployments now work

PR #287, opened 2017-10-03
docs/internal: spec for GET /tenants/:id/limits/storage

bboozzoo:
Fixes: [MEN #1462]

The origianl ticket mentioned `/limits/tenants/:id/storage`, I made that `/tenants/:id/limits/storage` to start with a per tenant endpoints namespace.

@mendersoftware/rndity

### DIFF
--- a/docs/internal_api.yml
+++ b/docs/internal_api.yml
@@ -1,0 +1,85 @@
+swagger: '2.0'
+info:
+  title: Deployments Internal API
+  version: '1'
+  description: |
+    Internal API of deployments service
+
+host: 'docker.mender.io'
+basePath: '/api/internal/v1/deployments'
+schemes:
+  - https
+
+responses:
+  NotFoundError: # 404
+    description: Not Found.
+    schema:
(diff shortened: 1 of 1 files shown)


mchalski:
won't it be a problem in implementation though?

the other version was more in line with the resource organization in `deployments`. if we add `limits/tenants/:id/storage` it naturally falls under a new `limits` resource (with all its controllers, views, etc). btw this new resource is already created in another pr.

what would it look like with `tenants/:id/limits` - a new `tenants` resource?

bboozzoo:
The idea was that we're going to have more `/tenant/:id` endpoints, for operations like triggering DB setup. There will likely fall under `/tenant/:id`, so the namespace can be reused. Besides `/limits/tenants/:id/` feels a bit awkward, like there's a limits resource that owns `tenant/:id/`, while with `/tenants/:id/..` the relation is more like 'tenant owns a limit'. Anyways, both work for me, if you prefer the original `/limits/..`, I'll do an update.

PR #289, opened 2017-10-17
S3 link expire time changes

bboozzoo:
@mendersoftware/rndity

### DIFF
--- a/docs/management_api.yml
+++ b/docs/management_api.yml
@@ -567,15 +567,6 @@ paths:
           description: Artifact identifier.
           required: true
           type: string
-        - name: expire
-          in: query
-          description: |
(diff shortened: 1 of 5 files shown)


docs/management_api.yml
```diff
@@ -571,9 +571,11 @@ paths:
           in: query
```
maciejmrowiec: maybe we can completely drop the parameter from doc

bboozzoo: I think it'd be ok to remove it. Not an API expert though :)

bboozzoo: Removed from API docs. Simplified test cases to only verify that expire is ignored.

PR #294, opened 2017-10-24
extended device deployment status reporting

bboozzoo:
Introduce a wrapper for device deployment status updates - `deployments.DeviceDeploymentStatus`. Make sure that it's passed though the layers, down to DB.

@mendersoftware/rndity @oleorhagen


resources/deployments/device_deployment.go
```diff
@@ -37,6 +37,17 @@ const (
 	DeviceDeploymentStatusDecommissioned = "decommissioned"
 )
 
+// DeviceDeploymentStatus is a helper type for reporting status changes through
+// the layers
+type DeviceDeploymentStatus struct {
+	// status reported by device
+	Status string `valid:"required"`
+	// substate reported by device
+	SubState *string
```
maciejmrowiec: any length limit?

bboozzoo: Should be fixed now, `SubState` is limited to 200 characters.

PR #300, opened 2017-11-15
Clarify artifact ownership by prefixing tenant ID to s3 object, and by setting tenant_id tag

GregorioDiStefano:


GregorioDiStefano:
@mendersoftware/rndity 

Setting the tenant tags is a little tricky for multiple reasons: 

1) Using a Presigned URL for upload to S3 with Tags doesn't seem to be supported, this requires me to set the tags after the artifact has been uploaded

2) Minio does not support tagging, and setting a tag to file actually truncates the file, and replaces it with an XML file describing the tags

The easiest way to get around the second problem is by making tenant tagging an configurable option. There is no reason we would ever use minio in production, but in case we do, we would have to turn off object tagging.

config.go
```diff
@@ -35,6 +35,9 @@ const (
 	SettingAwsS3BucketDefault = "mender-artifact-storage"
 	SettingAwsURI             = SettingsAws + ".uri"
 
+	SettingSetTenantTag        = SettingsAws + ".set_tenant_tag"
+	SettingSetTenantTagDefault = false
```
maciejmrowiec: bool in go always defaults to false

maciejmrowiec:
missing adjustments to other functions in filestore, delete, exist etc etc all need to use modified object name

GregorioDiStefano:
Added one-time tool for migrating to new object format (sneak peak, will make it nicer)

GregorioDiStefano:
thanks, fixed

resources/images/s3/filestorage.go
```diff
@@ -46,13 +46,14 @@ const (
 // Data layer for file storage.
 // Implements model.FileStorage interface
 type SimpleStorageService struct {
-	client *s3.S3
-	bucket string
+	client         *s3.S3
+	bucket         string
+	setTenantToken bool
```
mchalski: `setTenantTag` would be more consistent with `SettingSetTenantTag`

resources/images/s3/filestorage.go
```diff
@@ -46,13 +46,14 @@ const (
...
 }
 
 // NewSimpleStorageServiceStatic create new S3 client model.
 // AWS authentication keys are automatically reloaded from env variables.
-func NewSimpleStorageServiceStatic(bucket, key, secret, region, token, uri string) (*SimpleStorageService, error) {
+func NewSimpleStorageServiceStatic(bucket, key, secret, region, token, uri string, setTokenTag bool) (*SimpleStorageService, error) {
```
mchalski: `setTokenTag` is yet another variation of `setTenantToken` and `SettingSetTenantTag`, unify to `set tenant tag`

artifact_tenant_token_migration.py
```diff
@@ -0,0 +1,56 @@
...
+
+import logging as l
+import boto3
+import sys
+import pymongo
+
+l.basicConfig(level=l.INFO)
+s3 = boto3.resource('s3')
+
+MC = pymongo.MongoClient("mongodb://172.19.0.5")
```
mchalski: accept the config via args/env, get rid of defaults, fail when not specified

artifact_tenant_token_migration.py
```diff
@@ -0,0 +1,56 @@
...
+    l.info("setting tags [%s] for: %s" % (new_tag, obj_key))    
+    S3_CLIENT.put_object_tagging(Bucket=BUCKET, Key=obj_key, Tagging=new_tag)
+
+
+def copy_to_new_obj(old_key, new_key):
+    l.info("copying object: %s to %s" % (old_key, new_key))
+    S3_CLIENT.copy_object(Bucket=BUCKET, CopySource="%s/%s" % (BUCKET, old_key), Key=new_key)
+    #b.delete_object(Bucket="BucketName", Key="OriginalName")
+
+for db in MC.database_names():
```
mchalski: wrap in main()? with arg parsing and all

artifact_tenant_token_migration.py
```diff
@@ -0,0 +1,56 @@
...
+    new_tag = {"TagSet" :[{"Key": "tenant_id", "Value": tenant}]}
+
+    l.info("setting tags [%s] for: %s" % (new_tag, obj_key))    
+    S3_CLIENT.put_object_tagging(Bucket=BUCKET, Key=obj_key, Tagging=new_tag)
+
+
+def copy_to_new_obj(old_key, new_key):
+    l.info("copying object: %s to %s" % (old_key, new_key))
+    S3_CLIENT.copy_object(Bucket=BUCKET, CopySource="%s/%s" % (BUCKET, old_key), Key=new_key)
+    #b.delete_object(Bucket="BucketName", Key="OriginalName")
```
mchalski: huh?

GregorioDiStefano:
updated with nice migration script + test (83% test coverage)

resources/images/s3/filestorage.go
```diff
@@ -204,13 +222,36 @@ func (s *SimpleStorageService) UploadArtifact(ctx context.Context,
 			"Artifact upload failed with HTTP status %v", resp.Status)
 	}
 
+	if id := identity.FromContext(ctx); len(id.Tenant) > 0 && s.setTenantTag {
```
mchalski: consider erroring out when 'set tenant tag' is set and there's no 'tenant' in the jwt

mchalski: consider erroring out when 'set tenant tag' is set and there's no 'tenant' in the jwt

resources/images/s3/filestorage.go
```diff
@@ -112,9 +117,19 @@ func NewSimpleStorageServiceDefaults(bucket, region string) (*SimpleStorageServi
 	}, nil
 }
 
+func getImageByTenant(ctx context.Context, objectID string) string {
+	// If multi-tenancy, prefix objectID with tenant id
+	if id := identity.FromContext(ctx); id != nil && len(id.Tenant) > 0 {
```
mchalski: also consider erroring out here. this code will work fine, but on the off chance that:
- the 'set tenant tag' is set
- there's no 'tenant' in the jwt

maybe it's better to fail.

resources/images/s3/filestorage.go
```diff
@@ -43,14 +46,15 @@ const (
 // Data layer for file storage.
 // Implements model.FileStorage interface
 type SimpleStorageService struct {
-	client *s3.S3
-	bucket string
+	client       *s3.S3
+	bucket       string
+	setTenantTag bool
```
mchalski: on second thought, maybe this could be renamed to `haveMultitenant` or `withMultitenant`. it'll surely come in handy for other MT-related logic (after my requests it'll already be used both when uploading an artifact, and getting one).

mchalski: PS. along with the associated setting in config

migrations/artifact_tenant_token_migration.py
```diff
@@ -0,0 +1,87 @@
...
+                self.set_s3_tenant_tag(image_id, tenant_token)
+                new_object_key = "%s/%s" % (tenant_token, image_id)
+
+                l.info("%s has an artifact: %s that should be renamed to: %s" % (db, image_id, new_object_key)) 
+                self.copy_to_new_obj(image_id, new_object_key)
+
+
+if __name__ == "__main__":
+    MC = pymongo.MongoClient(os.getenv("MONGODB_MENDER"))
+    BUCKET = "mender-artifacts-greg"
```
mchalski: env?

migrations/artifact_tenant_token_migration.py
```diff
@@ -0,0 +1,87 @@
...
+    S3_CLIENT = boto3.client('s3')
+    S3_BUCKET = s3.Bucket(BUCKET)
+    
+    migration = Migration(MC, S3_CLIENT, S3_BUCKET)
+    migration.set_tenant_databases()
```
mchalski: just a general note but not a blocker.

instead of making the Migration stateful, what about an alternative pattern?
- static method get_tenant_databases(), returns list
- `perform(databases)`

then you don't have to remember the `.set_tenant_databases`, without which the instance is unusable and it's easy to miss. stateful classes of course make sense, but not in this case imo.

with passing tenant dbs - the interface of the class is simple, you can't misuse it. it doesn't make much difference in a one-off tool, but in general I'd prefer it for the sake of clarity.

personally what I would do however is: embed the tenant db iteration within `perform()`. then it's even simpler. wrap it up in a helper for clarity but keep it private and don't access explicitly.

migrations/artifact_tenant_token_migration.py
mchalski: also, it's a public repo - Greg's bucket doesn't fit here ;)

GregorioDiStefano: i didnt want to expose our MongoDB host to the public, ill change the bucket to an env variable as well

config.go
```diff
@@ -117,5 +120,6 @@ var (
 		{Key: SettingDbSSL, Value: SettingDbSSLDefault},
 		{Key: SettingDbSSLSkipVerify, Value: SettingDbSSLSkipVerifyDefault},
 		{Key: SettingGateway, Value: SettingGatewayDefault},
+		{Key: SettingUsingMultiTenancy, Value: SettingSetTenantTagDefault},
```
mchalski: SettingSetTenantTagDefault -> SettingUsingMultiTenantcyDefault

resources/images/s3/filestorage.go
```diff
@@ -112,9 +117,29 @@ func NewSimpleStorageServiceDefaults(bucket, region string) (*SimpleStorageServi
 	}, nil
 }
 
+func getImageByTenant(ctx context.Context, objectID string, tenantExpected bool) (string, error) {
```
mchalski: how about making this a method on `(s *SimpleStorageService)`? then you get rid of the awkward `tenantExpected` thing, you already have this info in `s`.

ignore if there's a reason for keeping this separate.

config.go
```diff
@@ -25,8 +25,9 @@ const (
 	SettingHttpsCertificate = SettingHttps + ".certificate"
 	SettingHttpsKey         = SettingHttps + ".key"
 
-	SettingListen        = "listen"
-	SettingListenDefault = ":8080"
+	SettingListen            = "listen"
+	SettingListenDefault     = ":8080"
+	SettingUsingMultiTenancy = "using_multitenancy"
```
maciejmrowiec: Minor thing but we use `have_multitenant` already at he gateway and gui, it would be good to be consistent with naming.

migrations/artifact_tenant_token_migration.py
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/python3
```
maciejmrowiec: One time use script, not sure if this should be committed into this repo.

config.go
maciejmrowiec: From other hand, do we really need that setting? Shouldn't we be able to determine this from request context like we do in other places ?

GregorioDiStefano: There seems to be some instances where we have multi tenancy enabled, but the token is missing from the context. @mchalski can you fill in the details?

migrations/artifact_tenant_token_migration.py
GregorioDiStefano: I can drop it, and keep it locally. It was worth getting review though :)

config.go
mchalski: hmm, ok, there's two things here.

Maciek is right that we could get rid of this altogether, and just deduce MT / ST from the tenant token. As it stands, the setting just serves to explicitly state our intent and as a constraint for token token verification. so yeah, I missed the larger context here.

@GregorioDiStefano what I meant was - if we *do* have this setting, then we should enforce that the token comes in with a tenant id. if we get rid of it - this is irrelevant, and we go with what we get.

mchalski: ...so, my opinion - let's get rid of it if it's not too much effort. sorry @GregorioDiStefano I missed it, I know we're running up to ~30 request changes already...

resources/images/s3/filestorage.go
```diff
@@ -112,9 +114,18 @@ func NewSimpleStorageServiceDefaults(bucket, region string) (*SimpleStorageServi
 	}, nil
 }
 
+func getImageByTenant(ctx context.Context, objectID string) string {
```
mchalski: to be consistent with the rest of the codebase: `s/image/artifact`

GregorioDiStefano:
The original PR included an option to enable/disable artifact tagging, which is needed, because minio breaks artifacts if you try tagging them. I re-introduced that option.

config.go
```diff
@@ -28,12 +28,14 @@ const (
...
-	SettingAwsS3Bucket        = SettingsAws + ".bucket"
-	SettingAwsS3BucketDefault = "mender-artifact-storage"
-	SettingAwsURI             = SettingsAws + ".uri"
+	SettingsAws                   = "aws"
+	SettingAwsS3Region            = SettingsAws + ".region"
+	SettingAwsS3RegionDefault     = "us-east-1"
+	SettingAwsS3Bucket            = SettingsAws + ".bucket"
+	SettingAwsS3BucketDefault     = "mender-artifact-storage"
+	SettingAwsURI                 = SettingsAws + ".uri"
+	SettingsAwsTagArtifact        = SettingsAws + ".tag_artifact"
```
maciejmrowiec: @mchalski fyi needed to go back to configurable option, minio completely breaks when tagging is used (overwrites actual file with meta-data file) and testing infrastructure for MT use minio for PR builds

mchalski:
@GregorioDiStefano @maciejmrowiec this stuff breaks minio, I propose adding another env var:
`DEPLOYMENTS_STORAGE_BACKEND = "S3" | "MINIO"`

ands set it accordingly. this is a pretty big change, but can't see a way around it:
- all compose files, including tests
- prod template
- prod template docs

anything else?

config.go
mchalski: yup, I'm online with Greg about it, thx.

PR #306, opened 2018-01-22
MEN-1724: add deployment device count and filtering by date

GregorioDiStefano:


GregorioDiStefano:
Preview before I start adding tests

resources/deployments/controller/controller_deployments.go
```diff
@@ -15,15 +15,17 @@
...
-	"net/http"
-	"net/url"
-
 	"github.com/ant0ine/go-json-rest/rest"
 	"github.com/asaskevich/govalidator"
 	"github.com/mendersoftware/go-lib-micro/identity"
 	"github.com/mendersoftware/go-lib-micro/log"
 	"github.com/mendersoftware/go-lib-micro/rest_utils"
 	"github.com/pkg/errors"
+	"net/http"
```
kjaskiewiczz: can you use previous import formatting (goimports like)?

resources/deployments/controller/controller_deployments.go
```diff
@@ -341,6 +369,18 @@ func (d *DeploymentsController) LookupDeployment(w rest.ResponseWriter, r *rest.
...
+	for idx, dep := range deps {
+		//REVIEW: should I return here on error?
+
+		deviceDeployments, err := d.model.GetDeviceStatusesForDeployment(ctx, *dep.Id)
+		if err != nil {
+			d.view.RenderError(w, r, err, http.StatusBadRequest, l)
+			return
+		}
+
+		deps[idx].Count = len(deviceDeployments)
```
kjaskiewiczz: maybe there is a better way of counting objects then retrieving all of them (there is a count in mongo, but I don't know how it works)

GregorioDiStefano:
To summarize the changes: 

1. the deployments/deployments/ endpoint has two new query parameters: created_after, and created_before, both which take a UTC epoch timestamp and filter the deployment history accordingly. 

2.  a device_count is added to every deployment listening, so you know how many devices were targeted with the deployment. 

3. an internal api is added, when you can specify a tenant_id and, and get it's deployment history. This requires a change in the go-lib-micro, that is simple, but I haven't pushed yet in case there is a better way of doing this

resources/tenants/controller/tenants.go
```diff
@@ -15,25 +15,37 @@
...
+	"net/http"
 )
 
 type Controller struct {
 	model model.Model
 }
 
+type DeploymentsControllerWithTenant struct {
+	model                 model.Model
+	deploymentsController deploymentsController.DeploymentsController
```
kjaskiewiczz: tenant controller with deployments controller?
is this the only way?
it looks really weird to me

GregorioDiStefano: I can give this a shot:

https://docs.mongodb.com/manual/reference/method/db.collection.count/

resources/deployments/controller/controller_deployments.go
GregorioDiStefano: can give this a shot:

https://docs.mongodb.com/manual/reference/method/db.collection.count/

resources/deployments/controller/controller_deployments.go
```diff
@@ -297,6 +299,24 @@ func ParseLookupQuery(vals url.Values) (deployments.Query, error) {
 		query.SearchText = search
 	}
 
+	createdBefore := vals.Get("created_before")
+	if createdBefore != "" {
+		if createdBeforeTime, err := parseEpochToTimestamp(createdBefore); err != nil {
+			return query, err
```
maciejmrowiec: error message won't contain from which param issue comes from, use errors.Wrap to deal with messaging

resources/deployments/controller/controller_deployments.go
```diff
@@ -341,6 +369,18 @@ func (d *DeploymentsController) LookupDeployment(w rest.ResponseWriter, r *rest.
 		return
 	}
 
+	for idx, dep := range deps {
```
maciejmrowiec: This is business logic, should be part of business model. Controller handles protocol logic.

resources/deployments/controller/controller_deployments.go
maciejmrowiec: Don't count it ourself, use count query.

maciejmrowiec: Alternatively, cache the count then creating deployment object in the first place. This way you avoid having to extra query

resources/deployments/deployment.go
```diff
@@ -80,6 +80,9 @@ type Deployment struct {
 	// Initialized with the "pending" counter set to total device count for deployment.
 	// Individual counter incremented/decremented according to device status updates.
 	Stats map[string]int `json:"-"`
+
+	// Total number of devices targeted
+	Count int `json:"device_count,omitempty"`
```
maciejmrowiec: This field is going to be stored in database, even if in current implementation is only alias for another query in response. Should be excluded from being stored  `bson:"-"`

maciejmrowiec: Call variable DeviceCount, Count is not self explanatory  on it's own.

resources/tenants/controller/tenants.go
maciejmrowiec: shouldn't that be model ?

resources/tenants/controller/tenants.go
```diff
@@ -54,3 +66,23 @@ func (c *Controller) ProvisionTenantsHandler(w rest.ResponseWriter, r *rest.Requ
...
+		rest_utils.RestErrWithLog(w, r, l, nil, http.StatusBadRequest)
+	}
+
+	//REVIEW: WithoutAuthContext would need to be added to go-lib-micro
+	tenantCtx := identity.WithoutAuthContext(ctx, &identity.Identity{
+		Tenant: tenantID,
+	})
+
+	r.Request = r.WithContext(tenantCtx)
+	c.deploymentsController.LookupDeployment(w, r)
```
maciejmrowiec: share model instead model (once business logic is cleaned up and moved down the stack)

maciejmrowiec:
API spec missing

resources/tenants/controller/tenants.go
GregorioDiStefano: but the controller is already doing all the query parsing, I'd have to re-implement it here .

GregorioDiStefano:
@maciejmrowiec can you take a look?
I think adding an acceptance test for the internal API makes sense, and I'll open a separate task.

resources/deployments/controller/controller_deployments.go
```diff
@@ -297,6 +298,24 @@ func ParseLookupQuery(vals url.Values) (deployments.Query, error) {
 		query.SearchText = search
 	}
 
+	createdBefore := vals.Get("created_before")
+	if createdBefore != "" {
+		if createdBeforeTime, err := parseEpochToTimestamp(createdBefore); err != nil {
+			return query, errors.Wrap(err, "timestamp parsing failed")
```
maciejmrowiec: include which query param are you parsing in error message, if you use both in request you'll need to know which is wrong

docs/internal_api.yml
```diff
@@ -99,6 +99,73 @@ paths:
...
+        - name: page
+          in: query
+          description: Results page number
+          required: false
+          type: number
+          format: integer
+          default: 1
+        - name: created_before
+          in: query
+          description: List only deployments created before Unix timestamp (UTC)
```
maciejmrowiec: db queries include greater or equal and less than equal, include that in docs

resources/tenants/controller/tenants.go
```diff
@@ -18,22 +18,34 @@ import (
...
+	model deploymentsModel.DeploymentsModel
+}
+
 func NewController(model model.Model) *Controller {
 	return &Controller{
 		model: model,
 	}
 }
 
+func NewDeploymentsTenantController(model deploymentsModel.DeploymentsModel) *TenantDeploymentController {
```
maciejmrowiec: Why not use existing Controller instead adding another one.

GregorioDiStefano: It's not possible to create a new method in a non-local package (I cant create a method of deploymentsController inside the 'tenant' package)

maciejmrowiec: Controller is in same package, it's in the same file.

GregorioDiStefano: right, I see what you mean now, and fixed it.

resources/tenants/controller/tenants.go
```diff
@@ -34,6 +37,12 @@ func NewController(model model.Model) *Controller {
 	}
 }
 
+func NewDeploymentController(depsModel *deploymentsModel.DeploymentsModel) *Controller {
```
maciejmrowiec: There is already New function for this structure, it should handle all the init handling.
Depending on which was used to create object some methods are functional and some not is asking for trouble. When creating Controller instance all methods should be usable.

Make NewController take both models as configuration.

routing.go
```diff
@@ -247,3 +250,13 @@ func NewTenantsResourceRoutes(controller *tenantsController.Controller) []*rest.
 		rest.Post(ApiUrlInternal+"/tenants", controller.ProvisionTenantsHandler),
 	}
 }
+
+func NewTenantDeploymentsRoutes(controller *tenantsController.Controller) []*rest.Route {
```
maciejmrowiec: There is already router for `tenantsController` no need for yet another one.
Made sense if there was different controller around, but when using one then it's just extra code.

routing.go
```diff
@@ -162,17 +162,21 @@ func NewRouter(c config.ConfigReader) (rest.App, error) {
 		new(deploymentsView.DeploymentsView))
 	limitsController := limitsController.NewLimitsController(limitsModel,
 		new(view.RESTView))
+
+	tenantsDeploymentController := tenantsController.NewDeploymentController(deploymentModel)
```
maciejmrowiec: there is already instance of this object available `tenantsController`,  no need for another one

docs/internal_api.yml
```diff
@@ -99,6 +99,73 @@ paths:
           description: Internal server error.
           schema:
            $ref: "#/definitions/Error"
+  /deployments/{id}/deployments:
```
maciejmrowiec: did you mean `tenants/{id}....`?

GregorioDiStefano: no, I didn't. 

The end point is: api/internal/v1/deployments/<id>/deployments

routing.go
```diff
@@ -237,13 +238,13 @@ func NewLimitsResourceRoutes(controller *limitsController.LimitsController) []*r
...
-func NewTenantsResourceRoutes(controller *tenantsController.Controller) []*rest.Route {
-
+func TenantRoutes(controller *tenantsController.Controller) []*rest.Route {
 	if controller == nil {
 		return []*rest.Route{}
 	}
 
 	return []*rest.Route{
 		rest.Post(ApiUrlInternal+"/tenants", controller.ProvisionTenantsHandler),
+		rest.Get(ApiUrlInternal+"/:tenant/deployments", controller.DeploymentsPerTenantHandler),
```
maciejmrowiec: route looks a bit off, should that be something like:

internalAPI `/tenants/:tenant/deployments`

GregorioDiStefano:
@maciejmrowiec i think this merge-able now

PR #307, opened 2018-02-14
MEN-1775: make it possible to upload artifact to specific tenant 

GregorioDiStefano:
Signed-off-by: Gregorio Di Stefano <greg.distefano+github@gmail.com>

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,3 @@
+codecov:
+  ignore:
+    - "utils/testing/*"
--- a/docs/internal_api.yml
+++ b/docs/internal_api.yml
@@ -166,6 +166,54 @@ paths:
         400:
           $ref: "#/responses/InvalidRequestError"
 
+  /tenants/{id}/artifacts:
+    post:
+      summary: Upload mender artifact
+      description: |
+        Upload mender artifact to a specific tenant. Multipart request with meta and artifact.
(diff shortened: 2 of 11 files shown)


GregorioDiStefano:
@maciejmrowiec this makes it possible to upload an artifact to a specific tenant, the code is essentially a copy of whats in the controller. 

Docs are missing, ill add them tomorrow morning. 

Does this need a unit test? I dont think it does, since it's essentially what we already have.

maciejmrowiec:
@mendersoftware/rndity

maciejmrowiec:
at least unit test on controller would be nice to have

GregorioDiStefano:
@maciejmrowiec @kjaskiewiczz can someone take a look?
The test coverage went down because I refactored some reused code into utils/tests and its now counting as untested code (fixed by adding it the codecov ignore list, but needs to be merged for it to count..)

.codecov.yml
```diff
@@ -0,0 +1,3 @@
+codecov:
```
maciejmrowiec: will this append or overwrite main global config? (global config specify coverage policy)

maciejmrowiec: other trick is use `_test.go` files, tests don't count into golang coverage

maciejmrowiec:
otherwhise looks ok

PR #311, opened 2018-04-17
GET /releases 

mchalski:
https://tracker.mender.io/browse/MEN-1428

modifications: releases have no `id` or `modified_ts` since they're not real resources yet (aggregated on the fly).

btw i'm using the store layer straight from the controller, it saves some work. the app layer is not necessary yet, add when really required.


docs/management_api.yml
```diff
@@ -923,16 +1003,63 @@ definitions:
...
     required:
       - limit
       - usage
     example:
       application/json:
         limit: 1073741824
         usage: 536870912
-
+  Release:
+    description: Groups artifacts with the same release name into a single resource.
```
maciejmrowiec: Grouping more of conternal implementation, conceptually this is a stand alone resource.

mchalski: so..."Software release definition"?

maciejmrowiec: Maybe something along the lines:

Release is a group of mender artifacts containing same version of software that targets variety of different platform types.

mchalski: fixed, but used `...different device types` instead of '...platform types' to be 1:1 with field names

PR #314, opened 2018-07-13
Update documentation of possible statuses of devices being part of de…

maciejmrowiec:
…ployment.

Changelog: none


maciejmrowiec:
User pointed out that docs are off, docs fix after verifying.

oleorhagen:
The way mender-artifact is set up now, a go-get will not install the tooling. Thus an explicit call to go install is required after go get. Therefore add: ```
go install github.com/mendersoftware/mender-artifact/cli/mender-artifact ```
and it will build just fine ;)

mchalski:
closing, I pulled those changes here https://github.com/mendersoftware/deployments/pull/316

PR #316, opened 2018-08-01
migrate to globalsign/mgo + master test fixes

mchalski:
https://tracker.mender.io/browse/MEN-2029

i'm pulling some extra fixes for broken acceptance tests from https://github.com/mendersoftware/deployments/pull/314.

'my' commits are actually the last few.

i'm also pinning bravado==9.2.2 because recent releases break everything.


kacf:
Back to the good old spurious failure now: Server returns wrong status code in some cases. Force merging.

PR #319, opened 2018-09-10
artifact upload: handle specific failures

kjaskiewiczz:
Issues: https://tracker.mender.io/browse/MEN-2012

I know this is a very simple and cheap solution, but I don't know if we can do much more here.

### DIFF
--- a/resources/images/controller/images.go
+++ b/resources/images/controller/images.go
@@ -21,6 +21,7 @@ import (
 	"mime/multipart"
 	"net/http"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/ant0ine/go-json-rest/rest"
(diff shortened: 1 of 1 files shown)


resources/images/controller/images.go
```diff
@@ -234,16 +235,37 @@ func (s *SoftwareImagesController) NewImage(w rest.ResponseWriter, r *rest.Reque
...
+
+	// handle specific cases
+
+	if strings.Contains(errMsg, "invalid checksum") {
+		return errors.New(errMsg[strings.Index(errMsg, "invalid checksum"):])
+	}
+
+	if strings.Contains(errMsg, "unsupported version") {
+		return errors.New(errMsg[strings.Index(errMsg, "unsupported version"):] +
+			"; supported versions are: 1, 2, 3")
```
kacf: Only 1 and 2 are supported ATM. 3 is WIP.

kjaskiewiczz: fixed

PR #326, opened 2019-01-02
Changed the Dockerfile to handle the build of the deployments service.

oleorhagen:
Changelog: The Dockerfile has changed so that all the source files are copied
into the container, then built using go install. This change was warranted by
the need for consistent builds across all services, and seemed to me the easiest
way to handle the cgo and musl dependencies consistently.

Signed-off-by: Ole Petter <ole.orhagen@cfengine.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,20 @@
-FROM alpine:3.4
-
+FROM golang:1.11.4-alpine3.8 as builder
 RUN apk update && apk upgrade && \
-     apk add ca-certificates && \
-     rm -rf /var/cache/apk/*
+     apk add \
+     xz-dev \
+     musl-dev \
+     gcc
+RUN mkdir -p /go/src/github.com/mendersoftware/deployments
+COPY . /go/src/github.com/mendersoftware/deployments
+RUN cd /go/src/github.com/mendersoftware/deployments && env CGO_ENABLED=1 go build
 
-RUN mkdir /etc/deployments
-
+FROM alpine:3.6
+RUN apk update && apk upgrade && \
+     apk add --no-cache ca-certificates
+RUN mkdir -p /etc/deployments
 EXPOSE 8080
-
-COPY ./config.yaml /etc/deployments/
-
-ENTRYPOINT ["/entrypoint.sh", "--config", "/etc/deployments/config.yaml"]
-
+COPY ./config.yaml /etc/deployments
 COPY ./entrypoint.sh /entrypoint.sh
-COPY ./deployments /usr/bin/
+COPY --from=builder /go/src/github.com/mendersoftware/deployments/deployments /usr/bin
+CMD ["./entrypoint.sh"]  
+ENTRYPOINT ["/entrypoint.sh", "--config", "/etc/deployments/config.yaml"]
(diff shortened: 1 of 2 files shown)


oleorhagen:
I guess the source files could also be removed after the binary is build. How slim do we want the container to be?

Dockerfile
```diff
@@ -1,16 +1,21 @@
-FROM alpine:3.4
+FROM golang:1.11-alpine
```
oleorhagen: Are we sensitive to alpine versions?

Dockerfile
```diff
@@ -1,16 +1,21 @@
...
 RUN apk update && apk upgrade && \
-     apk add ca-certificates && \
+     apk add ca-certificates \
+     xz-dev \
+     musl-dev \
+     gcc && \
      rm -rf /var/cache/apk/*
 
-RUN mkdir /etc/deployments
+RUN mkdir -p /go/src/github.com/mendersoftware/deployments
```
kacf: I would keep the original steps regarding `/etc` and `config.yaml`, and add a step to delete the source at the end. The fact that we build it inside the Dockerfile is an implementation detail, and the source is useless in the final container.

Dockerfile
```diff
@@ -1,16 +1,21 @@
...
 
 EXPOSE 8080
 
-COPY ./config.yaml /etc/deployments/
+COPY . /go/src/github.com/mendersoftware/deployments
+RUN cd /go/src/github.com/mendersoftware/deployments && go install
```
kacf: I would use `env CGO_ENABLED=1 go install`, just to be explicit, and to have it fail if it's missing or broken. Building without CGO will be something for local development only.

Dockerfile
kacf: Hmm, they have broken compatibility before... I think I would vote for trying to stay on a stable version (you can upgrade it if needed though), unless it's a lot of trouble.

Dockerfile
```diff
@@ -1,16 +1,21 @@
...
 
-ENTRYPOINT ["/entrypoint.sh", "--config", "/etc/deployments/config.yaml"]
+RUN apk del gcc
```
kacf: You can delete the two dev libraries as well.

Dockerfile
```diff
@@ -1,16 +1,21 @@
...
+
+ENTRYPOINT ["/entrypoint.sh", "--config", "/go/src/github.com/mendersoftware/deployments/config.yaml"]
 
 COPY ./entrypoint.sh /entrypoint.sh
-COPY ./deployments /usr/bin/
```
kacf: Does it work without this? I think `go install` installs to `/go/bin`, not `/usr/bin`, so I would expect this step (or a modified one) to still be needed.

oleorhagen: In the image `GOPATH` is in the `PATH` already, so this is not a problem

oleorhagen: Besides, this used to copy the static binary into the image, which is no longer how we are doing things if these changes are approved.

kacf:
Still working on getting [the client image updated](https://mender-jenkins.mender.io/job/mender-builder/2254/).

kacf:
It's updated now, so I restarted this job.

kacf:
> imo in this case it would be better to use docker multistage build - https://docs.docker.com/develop/develop-images/multistage-build/

+1

Didn't know about that feature!

kacf:
@kjaskiewiczz: Ping

PR #336, opened 2019-05-03
Added support for docker secrets

nielsavonds:
Adds support for providing S3 access key and secret using docker secrets.


oleorhagen:
@tranchitella it looks to me as no-one has really reviewed this. Would you mind having a look, and see if this is still valid? :pray:

PR #342, opened 2019-08-07
Add s3 credential validation

annymsMthd:
Signed-off-by: Joshua Benjamin <joshuabnjmn@gmail.com>


annymsMthd:
Had a little issue using an s3 bucket. It was because the Uri was left blank, which should be okay. I added some validation around the permissions of the aws account and removed CreateBucket from that flow. I'm not sure if this is the way you guys want it done but i figured I through this out there.

oleorhagen:
@tranchitella another one in which you have to decide upon the validity of :pray:

PR #344, opened 2019-09-22
MEN-2019 index deployments database 

merlin-northern:


merlin-northern:
for some reason I could not test it, will try again.

merlin-northern:
> for some reason I could not test it, will try again.

tested: the indices seem to be created.

store/mongo/migration_1_2_2.go
```diff
@@ -0,0 +1,82 @@
...
+		StatusIndexes,
+		DeviceIDStatusIndexes,
+		DeploymentIdIndexes,
+	}
+	for _, index := range indicesArrays {
+		// DropIndex will use the same rules for exploding the index name
+		// as EnsureIndexKey previously used to create the 'long' index
+		err := s.DB(m.db).
+			C(CollectionDevices).
+			DropIndex(index...)
```
kjaskiewiczz: do we need to drop old indexes? will there be some conflicts if we don't?

store/mongo/migration_1_2_2.go
```diff
@@ -0,0 +1,82 @@
...
+		DeploymentDeviceStatusPendingIndex,
+		DeploymentDeviceStatusInstallingIndex,
+		DeploymentDeviceStatusFinishedIndex,
+	}
+	for _, index := range indicesArrays {
+		// DropIndex will use the same rules for exploding the index name
+		// as EnsureIndexKey previously used to create the 'long' index
+		err := s.DB(m.db).
+			C(CollectionDeployments).
+			DropIndex(index...)
```
kjaskiewiczz: same question - why we need to drop old indexes?

store/mongo/migration_1_2_2.go
merlin-northern: > do we need to drop old indexes? will there be some conflicts if we don't?

There are two reasons:
 - `EnsureIndex` will not actually even connect to server once it returned with success, either `DropIndex` or `ResetIndexCache` has to be used to really be sure that `EnsureIndex`
 - it was a copy-paste from the way we deal with existing index in deployments

I do not think there are any conflicts that I can see.

store/mongo/migration_1_2_2.go
merlin-northern: > same question - why we need to drop old indexes?

same comment as above. I am not defending that code with all my strength, but as a precaution I would keep it.

store/mongo/migration_1_2_2.go
kjaskiewiczz: about `EnsureIndex`:
following requests **for the same index** will not contact the server
I think it's not the caseDropIndex

in the previous migration, `DropIndex` was used because we actually wanted to drop the index

merlin-northern: > about `EnsureIndex`:
> following requests **for the same index** will not contact the server
> I think it's not the case
> 
> in the previous migration, `DropIndex` was used because we actually wanted to drop the index

yes, of course, for the same index. it is only an extra precaution here.

store/mongo/migration_1_2_2_test.go
```diff
@@ -0,0 +1,118 @@
...
+		m := migrate.SimpleMigrator{
+			Session:     s,
+			Db:          tc.db,
+			Automigrate: true,
+		}
+
+		err := m.Apply(context.Background(), migrate.MakeVersion(1, 2, 2), migrations)
+		assert.NoError(t, err)
+
+		// devicesCollectionIndicesNames := []string{
```
kjaskiewiczz: why it's all commented out?

merlin-northern: > why it's all commented out?

oh wow:) I just pushed commented-out code, sorry about that. [fixed](https://github.com/mendersoftware/deployments/pull/344/commits/2c5365fe5a4bc388bbccbc6806ab1358eab60ffc)

PR #345, opened 2019-10-03
MEN-2818 test_artifacts_new_bogus_data test fix

merlin-northern:
ChangeLog:title

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -218,7 +218,7 @@ func (d *Deployments) handleArtifact(ctx context.Context,
 	if err != nil {
 		pW.Close()
 		<-ch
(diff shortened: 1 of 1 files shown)


oleorhagen:
Doesn't this basically revert `362569ba913454cf2646a11f73ea7ee0869e2d72`, with: 
```
    Make sure the most specific error is wrapped, not the generic one.
    
    Otherwise it breaks EOF handling in mender-artifact, which expects the
    cause of the error to be EOF.

```

oleorhagen:
Reverted in: https://github.com/mendersoftware/deployments/pull/347

PR #347, opened 2019-10-04
Revert "Make sure the most specific error is wrapped, not the generic one."

kjaskiewiczz:
This reverts commit 362569ba913454cf2646a11f73ea7ee0869e2d72.

We are doing that kind of wrapping because there is no easy way to unwrap the error.
API handler is looking at the error.Cause so it will be lost if we'll change wrapping here.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -218,7 +218,7 @@ func (d *Deployments) handleArtifact(ctx context.Context,
 	if err != nil {
 		pW.Close()
 		<-ch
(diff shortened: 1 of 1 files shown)


lluiscampos:
I agree on the revert, but can you add the justification in the commit message? Just the same as you wrote in the PR description

```
We are doing that kind of wrapping because there is no easy way to unwrap the error.
API handler is looking at the error.Cause so it will be lost if we'll change wrapping here.
```

kjaskiewiczz:
@lluiscampos sure; done

PR #351, opened 2019-10-14
Bumped gocyclo's complexity roof from 15 to 20

oleorhagen:
Changelog: Title

Signed-off-by: Ole Petter <ole.orhagen@northern.tech>

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -111,8 +111,8 @@ test:static:
     - if [ -n "$(go fmt)" ]; then echo 'Code is not formatted with "go fmt"'; false; fi
     # Perform static code analysys
     - go vet `go list ./... | grep -v vendor`
-    # Fail builds when the cyclomatic complexity reaches 15 or more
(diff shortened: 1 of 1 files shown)


kacf:
For some background, we already did this in mender-artifact and have found it to be more "fair" at 20 than at 15. But would like input from the backend folks.

oleorhagen:
From @lluiscampos in https://github.com/mendersoftware/deployments-enterprise/pull/19
> Few comments:
> 
>     * It think it is preferred to merge from os to enterprise. Cherry-picking will lead to conflicts sooner or later.
> 
>     * This change does not affect integ tests, so please stop the big pipeline
> 
>     * Why are we relaxing the criteria of gocyclo? This affects overall QA and I think more people needs to be involved (namely: me)

oleorhagen:
Gocyclo is relaxed as gocyclo's limit of fifteeen seems to be too low for some regular code patterns

oleorhagen:
@kacf @kjaskiewiczz 
See above

lluiscampos:
> Gocyclo is relaxed as gocyclo's limit of fifteeen seems to be too low for some regular code patterns

Ok. Thanks for bringing me here to this PR

PR #352, opened 2019-10-14
Fix dummy typo in README

lluiscampos:


lluiscampos:
Introduced in #350.

Skipping the build

PR #359, opened 2019-10-25
MC-1144 Run migrations on startup like other services do

merlin-northern:
ChangeLog:title

### DIFF
--- a/main.go
+++ b/main.go
@@ -117,13 +117,11 @@ func cmdServer(args *cli.Context) error {
 			3)
 	}
 
-	if args.Bool("automigrate") {
-		err = mongo.Migrate(context.Background(), mongo.DbVersion, dbSession, true)
-		if err != nil {
-			return cli.NewExitError(
-				fmt.Sprintf("failed to run migrations: %v", err),
-				3)
-		}
+	err = mongo.Migrate(context.Background(), mongo.DbVersion, dbSession, args.Bool("automigrate"))
+	if err != nil {
+		return cli.NewExitError(
+			fmt.Sprintf("failed to run migrations: %v", err),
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -117,13 +117,11 @@ func cmdServer(args *cli.Context) error {
...
 	}
 
-	if args.Bool("automigrate") {
-		err = mongo.Migrate(context.Background(), mongo.DbVersion, dbSession, true)
-		if err != nil {
-			return cli.NewExitError(
-				fmt.Sprintf("failed to run migrations: %v", err),
-				3)
-		}
+	err = mongo.Migrate(context.Background(), mongo.DbVersion, dbSession, true)
```
kjaskiewiczz: this way we'll always run migrations (not only check db version if automigrate is off)
we still need to handle automigrate command
it should look like
if automigrate 
mongo.Migrate(context.Background(), mongo.DbVersion, dbSession, true)
else
mongo.Migrate(context.Background(), mongo.DbVersion, dbSession, false)

merlin-northern: > this way we'll always run migrations (not only check db version if automigrate is off)
> we still need to handle automigrate command
> it should look like
> if automigrate
> mongo.Migrate(context.Background(), mongo.DbVersion, dbSession, true)
> else
> mongo.Migrate(context.Background(), mongo.DbVersion, dbSession, false)

ah yes! for got the last argument. [fixed](https://github.com/mendersoftware/deployments/pull/359/commits/326c66a1679ac194d0e22ef098b38e07ccc890aa)

merlin-northern: the deployments have different db.Migrate thats why I missed that.

PR #361, opened 2019-10-25
Removed dead assignment in DeploymentsPerTenantHandler()

olehermanse:
`ctx` was assigned, then reassigned. No behavior difference,
just making the function slightly more efficient / easier to read.

Reported by LGTM:
https://lgtm.com/rules/1510373626114/


oleorhagen:
Welcome to the Mender project, @olehermanse ^_^

oleorhagen:
LGTM, I don't know this code though, so let's make sure @kjaskiewiczz agrees, only reason I could see for doing it this way, is if the original context needs to be cached.  I guess the tests will show us if that is true though.

PR #362, opened 2019-10-25
Handle error from http.NewRequest() in GetDeviceInventory()

olehermanse:
According to sources for NewRequest:

https://golang.org/src/net/http/request.go?s=28198:28299#L828

It can return error in case URL is bad, context is nil, or
HTTP method is bad. I guess here, the most likely failure
would be a badly formed URL.

I assume that this "never happens", because it hasn't been a
problem, but it seems more correct to handle the error, just
in case.

Reported by LGTM:
https://lgtm.com/rules/1510373626114/

### DIFF
--- a/integration/inventory.go
+++ b/integration/inventory.go
@@ -1,4 +1,4 @@
-// Copyright 2017 Northern.tech AS
+// Copyright 2019 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


lluiscampos:
> Missing a `Changelog:` tag.
> 
> This should have been caught automatically by the CI. But we are having an issue with it ATM.

There **is** a Changelog tag, sorry. We are good here.

oleorhagen:
Not really necessary to cover the path here imo. Good to go

PR #376, opened 2019-12-10
tests: don't cleanup 'config' db when doing mongo cleanup

kjaskiewiczz:
reason:

pymongo.errors.OperationFailure: Direct writes against
config.transactions cannot be performed using a transaction or on a
session.

introduced in 3.6 apparently - sessions are used by default, and
'config' is untouchable on a session.

https://jira.mongodb.org/browse/SERVER-35244

### DIFF
--- a/store/mongo/datastore_mongo.go
+++ b/store/mongo/datastore_mongo.go
@@ -573,11 +573,10 @@ func (db *DataStoreMongo) InsertImage(ctx context.Context, image *model.Software
 		return err
 	}
 
-	res, err := collImg.InsertOne(ctx, image)
+	_, err := collImg.InsertOne(ctx, image)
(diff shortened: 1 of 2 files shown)


store/mongo/datastore_mongo.go
```diff
@@ -573,11 +573,10 @@ func (db *DataStoreMongo) InsertImage(ctx context.Context, image *model.Software
 		return err
 	}
 
-	res, err := collImg.InsertOne(ctx, image)
+	_, err := collImg.InsertOne(ctx, image)
 	if err != nil {
 		return err
 	}
-	image.Id = res.InsertedID.(string)
```
alfrunes: Thanks for cleaning this up :+1:

PR #377, opened 2019-12-23
MEN-3026 HEALTHCHECK added

merlin-northern:
ChangeLog:title

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,11 +10,12 @@ RUN cd /go/src/github.com/mendersoftware/deployments && env CGO_ENABLED=1 go bui
 
 FROM alpine:3.6
 RUN apk update && apk upgrade && \
-     apk add --no-cache ca-certificates xz
+     apk add --no-cache ca-certificates xz curl
 RUN mkdir -p /etc/deployments
 EXPOSE 8080
 COPY ./config.yaml /etc/deployments
 COPY ./entrypoint.sh /entrypoint.sh
 COPY --from=builder /go/src/github.com/mendersoftware/deployments/deployments /usr/bin
 CMD ["./entrypoint.sh"]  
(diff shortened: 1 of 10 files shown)


merlin-northern:
> I suggest closing this PR. While I support the addition of health checks, I believe this should be implemented in a common way exposing a `/healthz` end-point.

so you do not like that [/health](https://github.com/mendersoftware/deployments/pull/377/commits/acc3e2dad410214098913c3ae3f0b503e3666726#diff-cb73cbccfb145f8a678bc1d0dc9b56fcR38) ?

tranchitellawfp:
> > I suggest closing this PR. While I support the addition of health checks, I believe this should be implemented in a common way exposing a `/healthz` end-point.
> 
> so you do not like that [/health](https://github.com/mendersoftware/deployments/pull/377/commits/acc3e2dad410214098913c3ae3f0b503e3666726#diff-cb73cbccfb145f8a678bc1d0dc9b56fcR38) ?

I don't like `/api/internal/v1/deployments`, but my comment was more about: we either do it in all the services in a "common way" or we don't do it in a single service.

merlin-northern:
> > > I suggest closing this PR. While I support the addition of health checks, I believe this should be implemented in a common way exposing a `/healthz` end-point.
> > 
> > 
> > so you do not like that [/health](https://github.com/mendersoftware/deployments/pull/377/commits/acc3e2dad410214098913c3ae3f0b503e3666726#diff-cb73cbccfb145f8a678bc1d0dc9b56fcR38) ?
> 
> I don't like `/api/internal/v1/deployments/health`, but my comment was more about: we either do it in all the services in a "common way" or we don't do it in a single service.

originally I planned to move it into lib-go-micro. and make the /health endpoint common to all services, but did not find a good way, perhaps I will look again. putting all that on a side, the implementation of health can be different, as in this example, what do you think about that particular one?

tranchitella:
I'm closing this PR, we already implemented health checks.

PR #378, opened 2019-12-26
Refactor NewImage and NewImageForTenantHandler

tranchitella:
NewImage and NewImageForTenantHandler overlap: the first is the
management API which requires the user to be authenticated via JWT
token, the latter is the internal end-point used to upload an artifact
for a given tenant without authentication.

Their implementations overlap, thus we refactor the functions to share
the common part.

changelog: none

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3
+FROM python:3.7.5
 
 RUN pip3 install --quiet bravado==9.2.2 pymongo==3.6.1 pytest-ordering==0.5 minio crypto pytest==3.10.1 twisted requests pyyaml tz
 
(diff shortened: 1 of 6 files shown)


tests/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3
+FROM python:3.7.5
```
lluiscampos: I would drop the patch version here: using 3.7 should be good enough.

tranchitella: Unluckily, it is not enough as the regression has been introduced in Python 3.7 as well.
Anyway, looks like minio python library version 5.0.6 includes the fix so depending on it should fix the issue without pinning Python 3.7.5. I'm giving it a try later today!

tranchitella:
Superseded by https://github.com/mendersoftware/deployments/pull/382

PR #386, opened 2020-01-22
Add GetArtifactUri and DeleteArtifactUri to start workflow calls

tranchitella:
To allow the create artifact worker to download and delete the artifact
from S3, add the presigned get and delete URI for the artifact as
parameters to the HTTP call which starts the create_artifact workflow.

Changelog: none

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -39,7 +39,8 @@ import (
 const (
 	ArtifactContentType = "application/vnd.mender-artifact"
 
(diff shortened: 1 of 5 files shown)


mchalski:
@tranchitella this looks good. you even implemented the 'delete' link, which I missed from my spec.
let's just give KJ a chance to have a look and we'll merge.

PR #387, opened 2020-01-28
Additional test for create artifact internal end-point

tranchitella:
Changelog: none


tranchitella:
Merging it without waiting for the integration test, as the change is unrelated to those tests.

mchalski:
@tranchitella yup, looks good

PR #391, opened 2020-02-11
Fix artifact update end-point (MEN-3160)

tranchitella:
Changelog: none

### DIFF
--- a/store/mongo/images_test.go
+++ b/store/mongo/images_test.go
@@ -1,4 +1,4 @@
-// Copyright 2019 Northern.tech AS
+// Copyright 2020 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


store/mongo/images_test.go
```diff
@@ -267,3 +267,45 @@ func TestIsArtifactUnique(t *testing.T) {
 	}
 
 }
+
+func TestArtifactUpdate(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping TestIsArtifactUnique in short mode.")
```
mzedel: small typo there ;) `TestArtifactUpdate`

tranchitella: Ops, you are right; ready to force-push as soon as I get the test results.

tranchitella:
> LGTM - except one tiny thing...
> the one thing that remains untested though would be editing an artifact that has been used in a deployment - but that would be outside of this ticket IMO (and more related to MEN-2967)

I believe we should only fix this specific regression now, and the unit tests is needed for this specific method. I wasn't aware of MEN-2967, we can work out a fix for that as well, feel free to assign it to me and let Eystein prioritize it. :)

PR #393, opened 2020-02-22
QA-149: Use pipelines templates 

lluiscampos:
Use of CI templates and related fixes.

Pay special attention to the compile changes in production Dockerfile.

Depends on https://github.com/mendersoftware/mendertesting/pull/43, commit [52548e8](https://github.com/mendersoftware/deployments/pull/393/commits/52548e8c498a6ac94dc2c3354d3562e8e47d2b87) will be removed before merge.


lluiscampos:
This repo needs rework. We should not include commit [d82366d](https://github.com/mendersoftware/deployments/pull/393/commits/d82366d6d27693d3e254951da41dc689ee1c2b4d) as it would in effect remove lzma Artifact support from `deployments`. We need a custom template or a fix in the test runner.

However, shouldn't this have been caught by an integration test? I am understanding this wrong?

lluiscampos:
> While I'm curious why the different linking approach in the `Dockerfile` also works, I would consider it out of scope of this PR. 

Basically, I couldn't find a single integration test that uses lzma for the Artifact compression. I launched a separate PR for this: https://github.com/mendersoftware/integration/pull/820

I have to revert that change from this PR, I will try to find a good way to execute the binary from python:3 image, or else I will just don't use the template for acceptance tests here.

lluiscampos:
> I have to revert that change from this PR, I will try to find a good way to execute the binary from python:3 image, or else I will just don't use the template for acceptance tests here.

Fixed: dropped the old commit and fixed the test runner image instead. See last commit.

PR #399, opened 2020-03-11
Sirupsen -> sirupsen renaming

mzedel:
this is in preparation for a possible move to go modules/ align with the other repos/ reduce problems on certain filesystems...


tranchitella:
Superseded by https://github.com/mendersoftware/deployments/pull/432

mzedel:
🙏 thanks for taking care of this!

PR #406, opened 2020-04-06
Upload image streams the content directly to s3 while processing request

alfrunes:
Caching the artifact has the drawback of leaving the connection idle for
quite long, risking the connection to time out if the artifact is large.
This commit fixes that by parsing the form-data as a stream and
uploading the artifact as it is received.
As an added bonus: The progress bar in the UI is actually representative
of upload progress.

changelog: none


app/app.go
```diff
@@ -251,6 +263,25 @@ func (d *Deployments) handleArtifact(ctx context.Context,
 		return artifactID, uploadResponseErr
 	}
 
+	// Check that the submitted size is correct
+	var b [1]byte
+	N, err := multipartUploadMsg.ArtifactReader.Read(b[:])
+	if N > 0 && err == nil {
```
merlin-northern, line 269: I do not get this if, how is that checking the correctness of the size?

alfrunes: It's a bit cheeky, but the `LimitReader` returns `io.EOF` after the given number of bytes are read. Notice that this is the underlying reader the `LimitReader` is reading from. Thus, if we read more bytes from the reader, the actual payload is longer than the given size.

alfrunes:
I'm currently working on tuning the acceptance tests to the changes made to the endpoint parameters, but the implementation is finished.

PR #408, opened 2020-04-07
Deprecate form-data "size" in POST /artifacts requests

alfrunes:
This PR removes the size parameter from the POST request as it's considered redundant (we can't necessarily verify size ahead of time). The parameter is still allowed, but ignored.


s3/filestorage.go
```diff
@@ -219,63 +217,80 @@ func (s *SimpleStorageService) Exists(ctx context.Context, objectID string) (boo
...
+	//			},
+	//		},
+	//	}
+	//	if _, err := s.client.PutObjectTagging(input); err != nil {
+	//		l := log.FromContext(r.Context())
+	//		l.Warnf("failed to tag artifact : %s\n", objectID)
+	//	}
+	//}
+
+	//return nil
```
merlin-northern, line 293: this we do not need I presume?

api/http/api_deployments.go
```diff
@@ -402,42 +409,67 @@ func (d *DeploymentsApiHandlers) GenerateImage(w rest.ResponseWriter, r *rest.Re
...
-	multipartUploadMsg.MetaConstructor.Description = r.FormValue("description")
-
-	sizeValue := r.FormValue("size")
-	var size int64
-	var err error
-	if sizeValue != "" {
-		size, err = strconv.ParseInt(sizeValue, 10, 64)
+	var sizeParsed bool
+	for {
+		part, err := r.NextPart()
```
merlin-northern, line 418: I have never done that in go, so please excuse me asking: so we do nothing with the `part` thing?

s3/filestorage.go
alfrunes: Oops...

app/app.go
```diff
@@ -251,6 +263,25 @@ func (d *Deployments) handleArtifact(ctx context.Context,
...
+				l.Errorf(
+					"Failed to delete broken image %s: %s",
+					artifactID, err)
+			}
+		}()
+		return "", errors.Wrap(
+			ErrInvalidArtifactSize,
+			"size smaller than expected",
+		)
+	}
```
merlin-northern: discussed on slack

app/app.go
```diff
@@ -276,10 +248,17 @@ func (d *Deployments) handleArtifact(ctx context.Context,
...
-			ErrInvalidArtifactSize,
-			"size smaller than expected",
-		)
+		return "", ErrModelArtifactFileTooLarge
+	}
+	// close the pipe
+	pW.Close()
+
+	// Assign artifact size to the actual uploaded size
+	multipartUploadMsg.ArtifactSize = MaxImageSize - lr.N
```
merlin-northern, line 257: are we *sure* here that `MaxImageSize>=lr.N` ?

alfrunes: Yes, the `else if` above catches that case.

alfrunes:
Superseded by #406

PR #409, opened 2020-04-14
FIX: Artifact unique if device_types_compatible doesn't match

alfrunes:
Most likely caused by careless merging (my bad).

### DIFF
--- a/store/mongo/datastore_mongo.go
+++ b/store/mongo/datastore_mongo.go
@@ -676,7 +676,7 @@ func (db *DataStoreMongo) IsArtifactUnique(ctx context.Context,
 
 	query := bson.M{
 		"$and": []bson.M{
-			bson.M{
+			{
 				StorageKeyImageName: artifactName,
 			},
 			{
@@ -710,7 +710,6 @@ func (db *DataStoreMongo) IsArtifactUnique(ctx context.Context,
 		} else if len(i.ArtifactMeta.Depends) == 0 {
 			return false, nil
 		}
-		return false, err
 	}
 
 	return true, nil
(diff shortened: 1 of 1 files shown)


store/mongo/datastore_mongo.go
```diff
@@ -676,7 +676,7 @@ func (db *DataStoreMongo) IsArtifactUnique(ctx context.Context,
 
 	query := bson.M{
 		"$and": []bson.M{
-			bson.M{
+			{
```
kacf, line 679: To me the original looks like the correct one, but if it works...

store/mongo/datastore_mongo.go
```diff
@@ -710,7 +710,6 @@ func (db *DataStoreMongo) IsArtifactUnique(ctx context.Context,
 		} else if len(i.ArtifactMeta.Depends) == 0 {
 			return false, nil
 		}
-		return false, err
```
kacf, line 713: This code is a bit foreign to me, so apologies if I get this wrong, but I *think* this might not be correct. If you explicitly check for only one dependency count, that means that if you upload an artifact that has two dependencies, you essentially throw the whole uniqueness criterion out, and you can upload whatever you want. Including two artifacts that have identical depends information, as long as they have at least two entries.

For consistency I think we should apply the same constraints in the OS version as in the Enterprise version, even though we won't actually use the depends information. This could become important if people upgrade, for example.

store/mongo/datastore_mongo.go
alfrunes: I configured emacs to use gofmt with the `-s` option ("simplify"). It's exactly the same, but it helps making goreport happier :)

store/mongo/datastore_mongo.go
alfrunes: I know this looks like it violates the conflicting depends criterion, but this function is only used by the generate artifact endpoint where you can't specify any additional depends besides device type. It's not a pretty fix, but it's what's currently in the enterprise repo (that return line is just a glitch that I mistakenly made it in merging my last PR). We should probably follow up with creating a task adding an optional depends parameter to the generate artifact request and get this in the above function so it doesn't look that confusing.

kacf: Ah, that explains it!

PR #410, opened 2020-04-16
s3: Fix tenant-scoped tags on upload requests

alfrunes:


alfrunes:
@oleorhagen we don't have integration test coverage that covers this case and it shouldn't make a difference in practice.

PR #416, opened 2020-04-23
New deployments flow

kjaskiewiczz:
With the new deployments flow, deployments service is not creating device deployments during creation of the deployment. Instead, the deployments service saves the list of device IDs and when the device is asking for deployment, the deployments service is looking for the active (not finished) deployments that are newer than the one installed on the device.
If there are some deployments matching the filter, the deployments service check if the device is part of the deployment. If the device is part of the deployment, the deployment service creates a device deployment object and returns deployment instructions to the device.


kjaskiewiczz:
pipeline with backend integration tests fix: https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/139253475

PR #418, opened 2020-04-24
Snapshot on mongodb cursor is deprecated

tranchitella:
Instead of using the snapshot feature on mongodb cursor, sort the
documents by _id to process the records one by one.

Changelog: none


tranchitella:
@mzedel if you don't sort, you get the documents in the "natural order", which means that when you update a document it can end up to be moved at the end of the collection and the cursor will catch it up again.

PR #420, opened 2020-04-29
Rework migration 1.2.4 to run a single aggregation pipeline

tranchitella:


store/mongo/migration_1_2_4.go
```diff
@@ -50,6 +50,8 @@ func (m *migration_1_2_4) Up(from migrate.Version) error {
 	}
 	defer cur.Close(ctx)
 
+	allstats, _ := m.aggregateDeviceStatuses(ctx)
```
merlin-northern, line 53: either remove the error as return param, or use it here. I would check if there was an error here.

tranchitella: Ok, done!

PR #422, opened 2020-05-01
LookupDeployment returns total count

merlin-northern:
Towards reducing API calls required for paging.

 * X-Total-Count returned in header
 * app.LookupDeployment and db.Find return int64

ChangeLog:title
(cherry picked from commit 9a449b25d500f2513c302ae59111cc0fc147e444)

### DIFF
--- a/store/mongo/datastore_mongo.go
+++ b/store/mongo/datastore_mongo.go
@@ -1611,7 +1611,7 @@ func buildStatusKey(status string) string {
 }
 
 func (db *DataStoreMongo) Find(ctx context.Context,
-	match model.Query) ([]*model.Deployment, error) {
+	match model.Query) ([]*model.Deployment, int64, error) {
 
 	database := db.client.Database(mstore.DbFromContext(ctx, DatabaseName))
 	collDpl := database.Collection(CollectionDeployments)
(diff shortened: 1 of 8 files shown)


store/mongo/datastore_mongo.go
```diff
@@ -1680,16 +1680,20 @@ func (db *DataStoreMongo) Find(ctx context.Context,
 		options.SetLimit(int64(match.Limit))
 	}
 
+	count, err := collDpl.CountDocuments(ctx, query)
+	if err != nil {
+		count = -1
```
tranchitella, line 1685: @merlin-northern if count fails, return the error, I think it is the right thing to do

merlin-northern:
> Small request, but important (at least in my opinion): if count fails, better to return the error

sure. fixed. thanks.

store/mongo/datastore_mongo.go
merlin-northern: [my pleasure](https://github.com/mendersoftware/deployments/pull/422/commits/b93353ff381b40541cd3866a83589bafc9a8325f#diff-ea674f2e657b955202c2d74a4c7dec51R1685). fixed.

PR #423, opened 2020-05-02
Makefile added.

merlin-northern:
Targets:
 * all: default, calls test and build
 * build: calls `go build`
 * test: lists all modules and calls `go test`
 * clean: calls `go clean` and removes built exec file
 * run: calls build and ./service-name
 * run-automigrate: calls build and ./service-name server --automigrate

ChangeLog:title

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,25 @@
+SERVICE_NAME=deployments
+GO=go
+GOLIST=$(GO) list ./...
+GOBUILD=$(GO) build
+GOCLEAN=$(GO) clean
+GOTEST=$(GO) test
+
+all: test build
+
+build: 
+		CGO_ENABLED=1 $(GOBUILD) -o $(SERVICE_NAME)
+
+test: 
+		$(GOLIST) | grep -v /vendor | xargs -n1 -I {} -P 4 $(GOTEST) -v {}
+
+clean: 
+		$(GOCLEAN)
+		rm -f $(SERVICE_NAME)
+
+run: build
+		./$(SERVICE_NAME)
+
+run-automigrate: build
+		./$(SERVICE_NAME) server --automigrate
+


Makefile
```diff
@@ -0,0 +1,25 @@
...
+
+build: 
+		CGO_ENABLED=1 $(GOBUILD) -o $(SERVICE_NAME)
+
+test: 
+		$(GOLIST) | grep -v /vendor | xargs -n1 -I {} -P 4 $(GOTEST) -v {}
+
+clean: 
+		$(GOCLEAN)
+		rm -f $(SERVICE_NAME)
```
kacf, line 18: AFAIK, `go clean` already does this.

Makefile
```diff
@@ -0,0 +1,25 @@
...
+GO=go
+GOLIST=$(GO) list ./...
+GOBUILD=$(GO) build
+GOCLEAN=$(GO) clean
+GOTEST=$(GO) test
+
+all: test build
+
+build: 
+		CGO_ENABLED=1 $(GOBUILD) -o $(SERVICE_NAME)
```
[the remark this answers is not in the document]
kacf: +1

Makefile
```diff
@@ -0,0 +1,25 @@
+SERVICE_NAME=deployments
+GO=go
+GOLIST=$(GO) list ./...
+GOBUILD=$(GO) build
+GOCLEAN=$(GO) clean
+GOTEST=$(GO) test
+
+all: test build
```
kacf, line 8: I saw that @tranchitella had already done it like this in the workflows repository, but I would argue that it's more common that the first target (being the default if you only specify `make`) only builds. `all` could still do everything though, and the first target can be `default: build` or something like that. No strong feelings, but I thought I'd mention it, since it is how it is done in the client repositories and many "traditional" automake-based repositories and such.

tranchitella: @kacf sounds reasonable; the main purpose of my Makefile in workflows was simplifying my development workflow, thus I wanted it to test and, if it is successful, to build.

tranchitella:
@merlin-northern is this PR still actual? Or shall we close it?

merlin-northern:
> I suggest closing this PR

I think we decided to go with it, maybe lets just make the decision during next qa meeting, I would like to get it in, after I address all the comments.

tranchitellawfp:
> I think we decided to go with it, maybe lets just make the decision during next qa meeting, I would like to get it in, after I address all the comments.

Ack, so we either merge it (after fixes) or close it.

lluiscampos:
> > I suggest closing this PR
> 
> I think we decided to go with it, maybe lets just make the decision during next qa meeting, I would like to get it in, after I address all the comments.

@merlin-northern No need to discuss this again in a QA meeting. We agreed in that you would rework this PR with the consensus of the team, so that we can get it as the reference for the rest of the repos.

See the minutes of meeting for May meeting [here](https://docs.google.com/document/d/1neXnS16brsb7Jdn-Sg-f1Wyoyo_O5MJIAW7_FzTKpfY/edit#heading=h.o2nwfl1y8ly).

PR #424, opened 2020-05-04
s3: FIX part numbering starting from 1

alfrunes:
Currently the part numbering starts from 0 which is causing s3 to return
an API error trying to upload artifacts. Sadly minio and hence the
acceptance tests does not cover this part of the specification so this
bug slipped through.

### DIFF
--- a/s3/filestorage.go
+++ b/s3/filestorage.go
@@ -233,7 +233,7 @@ func (s *SimpleStorageService) uploadMultipart(
 	contentType string,
 ) error {
 	const maxPartNum = 10000
-	var partNum int64
+	var partNum int64 = 1
 	var rspUpload *s3.UploadPartOutput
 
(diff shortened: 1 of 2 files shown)


s3/filestorage.go
```diff
@@ -290,7 +290,7 @@ func (s *SimpleStorageService) uploadMultipart(
 
 	// The following is loop is very similar to io.Copy except the
 	// destination is the s3 bucket.
-	for partNum = 1; partNum < maxPartNum; partNum++ {
+	for partNum++; partNum < maxPartNum; partNum++ {
```
oleorhagen, line 293: This line struct me as a little odd. Is it supposed to start from two?

alfrunes: yes, the 1 index is already "used" above.

PR #429, opened 2020-05-08
Store DeviceCount value for Deployments in the database

tranchitella:
This commit does not include a migration; instead, when retrieving
deployments from the database, if no value is present the service
calculates it, querying and counting the number of DeviceDeployment
documents, and stores it in the document to avoid further queries to
retrieve this value.

Changelog: none

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -661,6 +661,10 @@ func (d *Deployments) GetDeployment(ctx context.Context,
 		return nil, errors.Wrap(err, "Searching for deployment by ID")
 	}
 
+	if err := d.setDeploymentDeviceCountIfUnset(ctx, deployment); err != nil {
+		return nil, err
+	}
+
 	return deployment, nil
 }
 
@@ -863,6 +867,10 @@ func (d *Deployments) createDeviceDeployment(ctx context.Context, deviceID strin
 
 	deviceDeployment.Created = deployment.Created
 
+	if err := d.setDeploymentDeviceCountIfUnset(ctx, deployment); err != nil {
+		return nil, err
+	}
+
 	if err := d.db.InsertDeviceDeployment(ctx, deviceDeployment); err != nil {
 		return nil, err
 	}
@@ -1048,6 +1056,22 @@ func (d *Deployments) GetDeviceStatusesForDeployment(ctx context.Context,
 	return statuses, nil
 }
 
+func (d *Deployments) setDeploymentDeviceCountIfUnset(ctx context.Context, deployment *model.Deployment) error {
+	if deployment.DeviceCount == nil {
+		deviceCount, err := d.db.DeviceCountByDeployment(ctx, *deployment.Id)
+		if err != nil {
+			return errors.Wrap(err, "counting device deployments")
+		}
+		err = d.db.SetDeploymentDeviceCount(ctx, *deployment.Id, deviceCount)
+		if err != nil {
+			return errors.Wrap(err, "setting the device count for the deployment")
+		}
+		deployment.DeviceCount = &deviceCount
+	}
+
+	return nil
+}
+
 func (d *Deployments) LookupDeployment(ctx context.Context,
 	query model.Query) ([]*model.Deployment, int64, error) {
 	list, totalCount, err := d.db.Find(ctx, query)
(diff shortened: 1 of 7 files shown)


app/app.go
```diff
@@ -1061,11 +1081,8 @@ func (d *Deployments) LookupDeployment(ctx context.Context,
 	}
 
 	for _, deployment := range list {
-		if deviceCount, err := d.db.DeviceCountByDeployment(ctx,
-			*deployment.Id); err != nil {
-			return nil, 0, errors.Wrap(err, "counting device deployments")
-		} else {
-			deployment.DeviceCount = deviceCount
+		if err := d.setDeploymentDeviceCountIfUnset(ctx, deployment); err != nil {
```
merlin-northern, line 1084: if multiple instances of deployments service call `setDeploymentDeviceCountIfUnset` at the same time, isn't it possible that [$inc](https://github.com/mendersoftware/deployments/pull/429/commits/01aa2db54de6e0b7f64891d5f1f0bfb9b0de895e#diff-ea674f2e657b955202c2d74a4c7dec51R1533) will have unwanted outcome?

mchalski: that would be my concern as well. maybe make it idempotent after all, and just set a value to X?

tranchitella: @merlin-northern @mchalski right, let's try a different approach: set if not set, inc (which is atomic) if already set. This should solve multiple instances working on the same deployment in a nice and clean way.

tranchitella:
> besides the `$inc` remark, it looks good, I like the no-migration pattern.
> 
> I think it'll cleanly port to `-enterprise` with dynamic deployments - am I correct?:
> 
> * set `DeviceCount` on `app.GetDeploymentForDeviceWithCurrent`
> * leave in the `setDeploymentDeviceCountIfUnset` to catch old deployments without `DeviceCount`
> * but also recalc `DeviceCount` for dynamic deployments on `app.GetNextDeploymentForDeviceWithCurrent`
> 
> I had some concerns about decommissioning, but double checked it's not a problem, this doesn't decrement the count (we had some unrelated hiccups with decommissioning/aborting, hence my caution).

@mchalski yes, it should be easily portable to deployments-enterprise, let's see what comes up with the auto-bot after merging into master. :)

store/mocks/DataStore.go
```diff
@@ -118,6 +118,34 @@ func (_m *DataStore) DeleteImage(ctx context.Context, id string) error {
 	return r0
 }
 
+// SetDeploymentDeviceCount provides a mock function with given fields: ctx, deploymentID, count
+func (_m *DataStore) SetDeploymentDeviceCount(ctx context.Context, deploymentID string, count int) error {
+	ret := _m.Called(ctx, deploymentID, count)
+
+	var r0 error
+	if rf, ok := ret.Get(1).(func(context.Context, string, int) error); ok {
```
merlin-northern, line 126: how can it work with `Get(1)` shouldn't it be `Get(0)`?

store/mocks/DataStore.go
```diff
@@ -118,6 +118,34 @@ func (_m *DataStore) DeleteImage(ctx context.Context, id string) error {
...
+		r0 = rf(ctx, deploymentID, count)
+	} else {
+		r0 = ret.Error(1)
```
merlin-northern, line 129: same here

store/mocks/DataStore.go
```diff
@@ -118,6 +118,34 @@ func (_m *DataStore) DeleteImage(ctx context.Context, id string) error {
...
+
+// IncrementDeploymentDeviceCount provides a mock function with given fields: ctx, deploymentID, increment
+func (_m *DataStore) IncrementDeploymentDeviceCount(ctx context.Context, deploymentID string, increment int) error {
+	ret := _m.Called(ctx, deploymentID, increment)
+
+	var r0 error
+	if rf, ok := ret.Get(1).(func(context.Context, string, int) error); ok {
+		r0 = rf(ctx, deploymentID, increment)
+	} else {
+		r0 = ret.Error(1)
```
merlin-northern, line 143: sirry, maybe I am already seeing things, but shouldn't it be index 0 here instead of 1

store/mocks/DataStore.go
tranchitella: https://github.com/mendersoftware/deployments/pull/431

store/mocks/DataStore.go
tranchitella: https://github.com/mendersoftware/deployments/pull/431

store/mocks/DataStore.go
tranchitella: https://github.com/mendersoftware/deployments/pull/431

PR #430, opened 2020-05-08
Add(stalebot): Add stalebot configuration

oleorhagen:
Changelog: None

Signed-off-by: Ole Petter <ole.orhagen@northern.tech>


oleorhagen:
Seems PR's are piling up, so why not?

oleorhagen:
Stopping pipelines as they should be un-influenced

PR #432, opened 2020-05-15
Upgrade vendorized go-lib-micro to latest version

tranchitella:
Changelog: none


tranchitella:
@merlin-northern yes, it is because of the capital letter. I've done it in two commits because it's how it has been handled in other cases (I used a PR from Manuel as example).

PR #433, opened 2020-05-15
device decommissioning fix

kjaskiewiczz:
Issue: https://tracker.mender.io/browse/MEN-3512

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -834,8 +834,8 @@ func (d *Deployments) getNewDeploymentForDevice(ctx context.Context,
 
 	//get deployments newer then last device deployment
 	//iterate over deployments and check if the device is part of the deployment or not
-	for skip := 0; true; skip++ {
-		deployments, err := d.db.FindNewerActiveDeployments(ctx, lastDeployment, skip, 1)
+	for skip := 0; true; skip += 100 {
+		deployments, err := d.db.FindNewerActiveDeployments(ctx, lastDeployment, skip, 100)
 		if err != nil {
 			return nil, nil, errors.Wrap(err,
 				"Failed to search for newer active deployments")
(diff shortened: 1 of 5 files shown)


app/app.go
```diff
@@ -1169,43 +1182,61 @@ func (d *Deployments) AbortDeployment(ctx context.Context, deploymentID string)
...
 
-	for _, deviceDeployment := range deviceDeployments {
-
-		stats, err := d.db.AggregateDeviceDeploymentByStatus(
-			ctx, *deviceDeployment.DeploymentId)
+	// get deployments newer then last device deployment
+	// iterate over deployments and check if the device is part of the deployment or not
+	// if the device is part of the deployment create new, decommisioned device deployment
+	for skip := 0; true; skip++ {
+		deployments, err := d.db.FindNewerActiveDeployments(ctx, lastDeployment, skip, 1)
```
tranchitella, line 1223: sorry for the silly question: is it really needed to run N queries, where N is the number of active deployments? wouldn't it work the same way using pagination with, let's say, 100 records per page? Not a big deal, but it's a (micro?) optimization if you have several tens (or hundreds) of active deployments.

kjaskiewiczz: yup; we can set a limit to 100 if that's the best number;
in the past, we were expecting tenants to have several, not several tens or hundreds, of active deployments, but it looks like it's not the case (after looking into hM data);
it would be good if everyone will take a look at the real hM data to have a general overview of how customers are using mender IMO

kjaskiewiczz: done

tranchitella: @kjaskiewiczz we have a few tenants with tens of active deployments (just verified), but we expect it to grow when they'll start using dynamic deployments, so better to reduce the number of queries. Thanks for addressing it!

PR #435, opened 2020-05-18
app: device decommissioning: update stats before recalculating deployment status

kjaskiewiczz:


kjaskiewiczz:
@tranchitella that was fast! :)

tranchitella:
@kjaskiewiczz we need it as soon as possible in enterprise :)

PR #437, opened 2020-05-22
MEN-3411 Inventory client backport

merlin-northern:
ChangeLog:none


merlin-northern:
Needed for the MEN-3411, I wanted to separate this backport from the implementation itself to stay as clean as possible.

.gitlab-ci.yml
```diff
@@ -22,7 +22,7 @@ include:
     file: '.gitlab-ci-github-status-updates.yml'
 
 test:unit:
-  image: golang:1.11
+  image: golang:1.13
```
tranchitella, line 25: I believe golang:1.13 is the default (see https://github.com/mendersoftware/mendertesting/blob/master/.gitlab-ci-check-golang-unittests.yml#L30), maybe it makes sense to stick to the default?

Dockerfile.acceptance-testing
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.11 as builder
+FROM golang:1.13 as builder
```
lluiscampos, line 1: Thanks for the update. The main `Dockerfile` was already using 1.13, so now tests and binary are using the same :+1:

main_test.go
```diff
@@ -36,10 +36,10 @@ func init() {
 
 	flag.BoolVar(&runAcceptanceTests, "acceptance-tests", false, "set flag when running acceptance tests")
 	flag.StringVar(&cliArgsRaw, "cli-args", "", "for passing urfave/cli args (single string) when golang flags are specified (avoids conflict)")
-	flag.Parse()
 }
 
 func TestRunMain(t *testing.T) {
+	flag.Parse()
```
oleorhagen, line 42: Why?

go.mod
```diff
@@ -0,0 +1,20 @@
+module github.com/mendersoftware/deployments
+
+go 1.13
+
+require (
+	github.com/ant0ine/go-json-rest v3.3.3-0.20170913041208-ebb33769ae01+incompatible
+	github.com/asaskevich/govalidator v0.0.0-20180720115003-f9ffefc3facf
```
oleorhagen, line 7: Question:
Why is this on such an old sha? And not a released version?

main_test.go
merlin-northern: part of the migration to go v.1.13, required.
e.g.: https://golang.org/src/flag/example_test.go
_"// execute, typically at the start of main (not init!):"_

go.mod
merlin-northern: believe it or not, but `TestUnmarshalLogMessage` fails with the latest. 
* github.com/asaskevich/govalidator v0.0.0-20200428143746-21a406dcc535 test fails
* github.com/asaskevich/govalidator v0.0.0-20180720115003-f9ffefc3facf test passes
just change in the dep made the test fail, it was not that easy to find.

model/device.go
```diff
@@ -0,0 +1,65 @@
...
+	//device's group name
+	Group string `json:"-" bson:"group,omitempty"`
+
+	CreatedTs time.Time `json:"-" bson:"created_ts,omitempty"`
+	//Timestamp of the last attribute update.
+	UpdatedTs time.Time `json:"updated_ts" bson:"updated_ts,omitempty"`
+}
+
+// wrapper for device attributes names and values
+type DeviceAttributes map[string]DeviceAttribute
```
alfrunes, line 61: Why do you need this? Doesn't it suffice to use `[]DeviceAttribute` ?

model/device.go
```diff
@@ -0,0 +1,65 @@
...
+	Value       interface{} `json:"value" bson:",omitempty"`
+	Scope       string      `json:"scope" bson:",omitempty"`
+}
+
+// Inventory device wrapper
+type InvDevice struct {
+	//system-generated device ID
+	ID string `json:"id" bson:"_id,omitempty"`
+
+	//a map of attributes names and their values.
```
alfrunes, line 33: This comment doesn't apply.

model/device.go
merlin-northern: it is just copied from the deployments-enterprise, shall I remove it?

model/device.go
```diff
@@ -0,0 +1,65 @@
...
+	//device's group name
+	Group string `json:"-" bson:"group,omitempty"`
+
+	CreatedTs time.Time `json:"-" bson:"created_ts,omitempty"`
+	//Timestamp of the last attribute update.
+	UpdatedTs time.Time `json:"updated_ts" bson:"updated_ts,omitempty"`
+}
+
+// Device wrapper
+type Device struct {
```
alfrunes, line 45: I can't see this definition being used anywhere.

model/device.go
merlin-northern: yeah, cool, but it still is from the ent counterpart. maybe lets leave it for the sake of next re-back-port-merge to enterprise?

model/device.go
merlin-northern: I understand, but keeping it here may make one less conflict with the merge to ent. do you want me to remove it?

model/device.go
alfrunes: I think you can remove this and the Device definition (if it's not being used anywhere), but maybe do it in a follow-up PR if this applies to the enterprise counterpart as well.

model/device.go
alfrunes: Keep it in, but I'd suggest to remove it in a follow up PR and propagate to enterprise as well.

merlin-northern:
> @merlin-northern ahh, I hit approve but didn't notice that vendor is gone, thought it was just an update.
> I was certain we decided to keep it, so not sure how to proceed - is it on purpose, am I missing something ? :)

I removed ./vendor, and recreated it with `go mod vendor` resulting is this commit:
https://github.com/mendersoftware/deployments/pull/437/commits/ce5f5b34afaebfd68294001e6a2102e376c0ee76
I guess you can call it heavily-trimmed vendor tree, since it now contains 1341 files, while the old one contained 1662. if you have objections we can postpone the migration to modules.

mchalski:
@merlin-northern sorry for the confusion, I see it now.

model/device.go
merlin-northern: https://github.com/mendersoftware/deployments/pull/439

model/device.go
merlin-northern: https://github.com/mendersoftware/deployments/pull/439

PR #442, opened 2020-06-08
MEN-3646: create a new index, devices(deviceid,created,status)

tranchitella:
Migration: 1.2.5; drop the rendundant index devices(deviceid, status,
created), create a new index devices(deviceid, created, status) which
allows us to select the oldest deployment for a deviceid with a given
status.

Changelog: none


tranchitella:
> +1
> Maybe we should change the default connection string with the hard-coded write-concern as default?

As discussed privately, the defaults (readPreference, readConcern and writeConcern) are safe in mongodb, thus we can relay on them.

PR #452, opened 2020-06-19
API spec fixes: schema and hostname

lluiscampos:
* [API spec] Consistently use host hosted.mender.io across all repos
* [API spec] Consistently use https schemes across all repos

### DIFF
--- a/docs/internal_api.yml
+++ b/docs/internal_api.yml
@@ -5,7 +5,7 @@ info:
   description: |
     Internal API of deployments service
 
-host: 'docker.mender.io'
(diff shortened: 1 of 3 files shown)


docs/internal_api.yml
```diff
@@ -5,7 +5,7 @@ info:
   description: |
     Internal API of deployments service
 
-host: 'docker.mender.io'
```
tranchitella, line 8: This API is not exposed on hosted mender, and it is considered internal.
I don'w know what the policy is for these end-points, but hosted.mender.io is wrong.

PR #458, opened 2020-07-03
MEN-2907: Adjust swagger spec to integrate with new API docs

alfrunes:
Corrects `example` definitions and adds JWT as a security definition.

### DIFF
--- a/docs/devices_api.yml
+++ b/docs/devices_api.yml
@@ -12,6 +12,20 @@ basePath: '/api/devices/v1/deployments'
 schemes:
   - https
 
+consumes:
+  - application/json
+produces:
+  - application/json
+
+securityDefinitions:
+  DeviceJWT:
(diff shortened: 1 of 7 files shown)


docs/devices_api.yml
```diff
@@ -180,9 +171,27 @@ definitions:
...
+      status:
+        type: string
+        enum:
+          - installing
+          - downloading
+          - rebooting
+          - success
+          - failure
+          - already-installed
+      substate:
```
tranchitella, line 188: Maybe we miss noartifact here?

alfrunes: Well spotted!
I was about to say yes, but inspecting the code this status is a special case which is reported exclusively by the backend.

tranchitella: thanks for checking!

PR #461, opened 2020-07-21
Fix incorrect version list and broken link definition.

kacf:
Changelog: None

Signed-off-by: Kristian Amlie <kristian.amlie@northern.tech>


kacf:
Cancelling integration pipeline, I assume this has no effect on it (but please correct me if I'm wrong).

PR #478, opened 2020-09-03
MEN-3920: filter by accepted devices when deploying to group

tranchitella:
Changelog: none

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -21,7 +21,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-	"github.com/satori/go.uuid"
+	uuid "github.com/satori/go.uuid"
 
 	"github.com/mendersoftware/go-lib-micro/identity"
 	"github.com/mendersoftware/go-lib-micro/log"
@@ -44,7 +44,10 @@ const (
 	DefaultImageGenerationLinkExpire = 7 * 24 * time.Hour
 	PerPageInventoryDevices          = 512
 	InventoryGroupScope              = "system"
+	InventoryIdentityScope           = "identity"
 	InventoryGroupAttributeName      = "group"
+	InventoryStatusAttributeName     = "status"
+	InventoryStatusAccepted          = "accepted"
 )
 
 // maximum image size is 10G
@@ -701,6 +704,12 @@ func (d *Deployments) CreateDeployment(ctx context.Context,
 					Type:      "$eq",
 					Value:     group,
 				},
+				{
+					Scope:     InventoryIdentityScope,
+					Attribute: InventoryStatusAttributeName,
+					Type:      "$eq",
+					Value:     InventoryStatusAccepted,
+				},
 			},
 		}
 		for {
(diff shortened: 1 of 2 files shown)


app/app_test.go
```diff
@@ -165,6 +165,53 @@ func TestDeploymentModelCreateDeployment(t *testing.T) {
...
+			InvDevices: []model.InvDevice{
+				{
+					ID: "b532b01a-9313-404f-8d19-e7fcbe5cc347",
+				},
+			},
+			TotalCount: 1,
+
+			OutputBody: true,
+		},
+		"ok with group, two pagese": {
```
merlin-northern, line 183: nitpicking: pagese -> pages, right?

app/app_test.go
```diff
@@ -165,6 +165,53 @@ func TestDeploymentModelCreateDeployment(t *testing.T) {
...
+				Name:         pointers.StringToPointer("group"),
+				ArtifactName: pointers.StringToPointer("App 123"),
+			},
+
+			InvDevices: []model.InvDevice{
+				{
+					ID: "b532b01a-9313-404f-8d19-e7fcbe5cc347",
+				},
+			},
+			TotalCount: 2,
```
merlin-northern, line 194: how can that be possible when total count is 2 and there is one device returned?

app/app_test.go
```diff
@@ -200,10 +247,53 @@ func TestDeploymentModelCreateDeployment(t *testing.T) {
...
+						},
+					},
+				).Return(testCase.InvDevices, testCase.TotalCount, testCase.SearchError)
+
+				if testCase.TotalCount > len(testCase.InvDevices) {
+					mockInventoryClient.On("Search", ctx,
+						"tenant_id",
+						model.SearchParams{
+							Page:    2,
+							PerPage: PerPageInventoryDevices,
```
merlin-northern, line 278: I do not fully understand, what is the problem with more than one page in search results? also: why we simulate two pages with PerPage equal PerPageInventoryDevices and not 1?

tranchitella: ok, made it a bit more explicit!

tranchitella: PerPageInventoryDevices is constant to 512, I do not want to write 512 devices in the test, and the test for the "next" page is simply: I've got less devices than the total, so providing one device instead of two is enough to trigger next page.

app/app.go
```diff
@@ -717,6 +726,9 @@ func (d *Deployments) CreateDeployment(ctx context.Context,
 				break
 			}
 			constructor.Devices = append(constructor.Devices, inventoryDevicesToDevicesIds(devices)...)
+			if len(constructor.Devices) == count {
+				break
+			}
```
merlin-northern, line 731: this is just an extra condition in case the first search returned all the devices we need.

app/app_test.go
merlin-northern: right! (I thought about setting PerPage=1). but it is fine I guess.

merlin-northern (approved):
lgtm

app/app.go
tranchitella: yes, but it also prevents an extra call to the inventory if we have all the devices we expect, already.

tranchitella:
Integration test failure is spurious, docker-compose related.

PR #483, opened 2020-09-10
MEN-3908: add missing documentation for deploy to group API end-point

tranchitella:
Changelog: none

### DIFF
--- a/docs/management_api.yml
+++ b/docs/management_api.yml
@@ -154,6 +154,53 @@ paths:
         500:
           $ref: "#/responses/InternalServerError"
 
+  /deployments/group/{name}:
+
+    post:
+      operationId: Create Deployment for a Group of Devices
+      tags:
+        - Management API
+      security:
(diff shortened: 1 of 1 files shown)


docs/management_api.yml
```diff
@@ -154,6 +154,53 @@ paths:
...
+        - application/json
+      responses:
+        201:
+          description: New deployment created.
+          headers:
+            Location:
+              description: URL of the newly created deployment.
+              type: string
+        400:
+          $ref: "#/responses/InvalidRequestError"
```
alfrunes, line 198: Maybe we should start adding 401 responses here as well?
(I can see it's also not documented for the other endpoints either).

alfrunes (approved):
LGTM

I only have a small followup comment.

docs/management_api.yml
tranchitella: You are right, they are missing, I've pushed a new commit to add them.

PR #485, opened 2020-09-17
MEN-2613: missing description fields in Deployment schemas

tranchitella:


docs/devices_api.yml
```diff
@@ -242,6 +242,7 @@ definitions:
     properties:
       messages:
         type: array
+        description: Array log entries of a deployment
```
merlin-northern, line 245: should this be "An array of .."?

docs/management_api.yml
```diff
@@ -391,6 +391,7 @@ paths:
               - "00a0c91e6-7dec-11d0-a765-f81d4faebf7"
           schema:
             type: array
+            description: List of device IDs
```
merlin-northern, line 394: "A list .." perhaps?

docs/management_api.yml
```diff
@@ -1212,10 +1213,9 @@ definitions:
             release name.
       artifacts:
         type: array
+        description: List of artifacts for this release.
```
merlin-northern, line 1227: "A list .." perhaps?

docs/management_api.yml
```diff
@@ -884,8 +884,10 @@ definitions:
     properties:
       name:
         type: string
+        description: Name of the deployment
```
merlin-northern, line 887: Maybe lets decide whether we use the articles (a/an/the) in the descriptions or not?

docs/devices_api.yml
tranchitella: Array of, thanks!

docs/management_api.yml
tranchitella: I have the impression we already decided to NOT use articles in descriptions, and more or less this is what we have all across the different specification files. What do you think?

docs/management_api.yml
tranchitella: See comment below.

docs/management_api.yml
tranchitella: See comment below.

tranchitella:
Stopped pipeline, useless.

docs/management_api.yml
merlin-northern: I have no problem with no articles as long as it is consistent. no strong opinions, since I am not a native english speaker :) Spotify seems to be using articles e.g.: [The Spotify ID for the artist](https://developer.spotify.com/documentation/web-api/reference/artists/get-artist/) and Slack as well: [A unique ID for the incident](https://api.slack.com/docs/slack-status#v2_0_0)

merlin-northern (approved):
so, since I trust you so much, lgtm!

PR #497, opened 2020-10-19
MEN-4051: document artifact_provides and artifact_depends in API docs

tranchitella:
Changelog: title


tranchitella:
Only test failed in integration is the mtls, rsa-flavor, we can ignore.

lluiscampos (approved):

PR #499, opened 2020-10-19
MEN-4051: document artifact_provides and artifact_depends in API docs

tranchitella:
Changelog: title


tranchitella:
Unit tests for mongo fail because of `Error parsing command line: unrecognised option '--nssize'`.
This is due to the change in the GitLab pipeline template, we can ignore.

lluiscampos (approved):

PR #510, opened 2020-11-30
api: Generate artifact doesn't load entire request upon parsing

alfrunes:
Parses the multipart request form fields sequentially without reading
the entire request into memory and disk.
I also added a new LimitedReader interface with a custom error so that
the s3 package aborts the ongoing multipart upload if the file is too large.

### DIFF
--- a/api/http/api_deployments.go
+++ b/api/http/api_deployments.go
@@ -406,20 +406,25 @@ func formatArtifactUploadError(err error) error {
 func (d *DeploymentsApiHandlers) GenerateImage(w rest.ResponseWriter, r *rest.Request) {
 	l := requestlog.GetRequestLogger(r)
 
-	err := r.ParseMultipartForm(DefaultMaxMetaSize)
+	formReader, err := r.MultipartReader()
 	if err != nil {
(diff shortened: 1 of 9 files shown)


vendor/github.com/asaskevich/govalidator/validator.go
```diff
@@ -323,12 +323,12 @@ func IsNull(str string) bool {
 
 // HasWhitespaceOnly checks the string only contains whitespace
 func HasWhitespaceOnly(str string) bool {
-    return len(str) > 0 && rxHasWhitespaceOnly.MatchString(str)
```
tranchitella, line 326: Why are you changing a single file in the vendor folder?

tranchitella (approved):
Approved, I have a question, though.

vendor/github.com/asaskevich/govalidator/validator.go
alfrunes: I was inspecting these sources, and I think my editor must have run `gofmt` and saved.

alfrunes: Re-vendored and force pushed.

PR #520, opened 2020-12-07
fix mongo migration index not found error

prashanthjbabu:
**Issue** : mender deployments fails to start
**RCA** :  the index not found error message by mongo contains the name of the index as well
The error message by mender during migration is 
`time="2020-12-07T05:40:51Z" level=error msg="migration from 0.0.0 to 1.2.1 failed: index not found with name [deploymentconstructor.name_text_deploymentconstructor.artifactname_text]" db=deployment_service file=migrator_simple.go func="migrate.(*SimpleMigrator).Apply" line=106`
**Fix** : use strings.HasPrefix to check the type of error message instead of equating it directly

### DIFF
--- a/store/mongo/migration_1_2_1.go
+++ b/store/mongo/migration_1_2_1.go
@@ -16,6 +16,7 @@ package mongo
 import (
 	"github.com/globalsign/mgo"
 	"github.com/mendersoftware/go-lib-micro/mongo/migrate"
+	"strings"
 )
 
 type migration_1_2_1 struct {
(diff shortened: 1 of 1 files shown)


prashanthjbabu:
@kacf would you be able to take a look at this?

kacf:
Thanks for the contribution. Both the [1.7.x branch](https://github.com/mendersoftware/deployments/pull/521) and the 1.8.x branch have been end-of-life for some time now. Please submit this to the master branch instead; we always submit there first and then cherry-pick afterwards.

Going forward, I think @alfrunes may be able to help. He deals more with the database side of things and should be able to help you.

prashanthjbabu:
@kacf  Thanks for your response . I don't think the issue exists in master branch. I'll close this then!

PR #521, opened 2020-12-07
fix mongo migration index not found error

prashanthjbabu:
**Issue** : mender deployments fails to start
**RCA** :  the index not found error message by mongo contains the name of the index as well
The error message by mender during migration is 
`time="2020-12-07T05:40:51Z" level=error msg="migration from 0.0.0 to 1.2.1 failed: index not found with name [deploymentconstructor.name_text_deploymentconstructor.artifactname_text]" db=deployment_service file=migrator_simple.go func="migrate.(*SimpleMigrator).Apply" line=106`
**Fix** : use strings.HasPrefix to check the type of error message instead of equating it directly

### DIFF
--- a/migrations/migration_1_2_1.go
+++ b/migrations/migration_1_2_1.go
@@ -16,6 +16,7 @@ package migrations
 import (
 	"github.com/globalsign/mgo"
 	"github.com/mendersoftware/go-lib-micro/mongo/migrate"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


prashanthjbabu:
@kacf  would you be able to take a look at this?

prashanthjbabu:
closing this due to https://github.com/mendersoftware/deployments/pull/520#issuecomment-741626600

PR #531, opened 2021-01-12
Migrate from govalidator to ozzo-validation

alfrunes:
This is part of MEN-4332 for redacting sensitive information from the logs. With ozzo-validation we guarantee that no validation errors will leak sensitive information. As this turned out to be quite a complex, I decided to split MEN-4332 into two parts for this repository - masking query parameters in a second PR.


tranchitella:
@alfrunes there are several failures in the acceptance tests, is this PR ready for review?

alfrunes:
Sorry, I forgot to check the acceptance tests before creating this PR. Luckily they picked up a quite significant error in my refactoring.
Should be all good now.

alfrunes:
The PR is ready for review now. Sorry about the mistake before.
(I'd recommend reviewing this one commit at the time as the first one is the most important).

model/device_deployment.go
```diff
@@ -17,119 +17,143 @@ package model
...
-func NewDeviceDeployment(deviceId, deploymentId string) (*DeviceDeployment, error) {
+func NewDeviceDeployment(deviceId, deploymentId string) *DeviceDeployment {
 
 	now := time.Now()
-	initStatus := DeviceDeploymentStatusPending
 
-	uid := uuid.NewV4()
+	uid, err := uuid.NewRandom()
+	if err != nil {
+		panic(errors.Wrap(err, "failed to generate random uuid (v4)"))
```
kjaskiewiczz, line 115: I'm not sure how it will be handled by the go-json-rest library
I think 500 will be sent, which is good, but without the body, which may not be good, because in all other cases we are sending an error message with http 500 responses

alfrunes: The only way this can happen is if the uuid library fails to read from the RNG (which I only think can happen if the machine runs out of entropy). Either way I don't think we should expose this information to the API and let the recovery middleware push a the stacktrace to the logs.

kjaskiewiczz (approved):
looks good

PR #538, opened 2021-01-18
MEN-4332: Remove sensitive information from logs

alfrunes:
Removes device type, artifact- and release name from log entries. I also did some minor lint-check fixes.

- [x] Blocked by mendersoftware/deployments-enterprise#291


api/http/api_deployments.go
```diff
@@ -119,9 +119,20 @@ func (d *DeploymentsApiHandlers) GetReleases(w rest.ResponseWriter, r *rest.Requ
...
 
-	name := r.URL.Query().Get("name")
+	q := r.URL.Query()
+	name := q.Get("name")
 
 	if name != "" {
+		defer func() {
+			var redacted bool
+			if q.Get("name") != "" {
+				q.Set("name", "REDACTED")
```
merlin-northern, line 129: I will keep repeating that, sorry, could we please have a const "REDACTED"?

api/http/api_deployments.go
```diff
@@ -965,8 +990,19 @@ func parseEpochToTimestamp(epoch string) (time.Time, error) {
 func (d *DeploymentsApiHandlers) LookupDeployment(w rest.ResponseWriter, r *rest.Request) {
 	ctx := r.Context()
 	l := requestlog.GetRequestLogger(r)
+	q := r.URL.Query()
+	defer func() {
+		var redacted bool
+		if search := q.Get("search"); search != "" {
+			q.Set("search", "REDACTED")
+			redacted = true
```
merlin-northern, line 997: why do we need the flag `redacted` here?

api/http/api_deployments.go
```diff
@@ -352,7 +352,6 @@ func (d *DeploymentsApiHandlers) newImageWithContext(ctx context.Context, w rest
 
 	// parse multipart message
 	multipartUploadMsg, err := d.ParseMultipart(formReader)
-	defer r.MultipartForm.RemoveAll()
```
merlin-northern, line 344: I am not sure about that, why are we dropping that?

merlin-northern (commented):
left some comments, not critical, I am not sure about one thing, otherwise looks good.

api/http/api_deployments.go
alfrunes: This statement is a NOOP. We're not calling `r.ParseMultipartForm` anymore.

api/http/api_deployments.go
alfrunes: We don't :)

merlin-northern (approved):
lgtm!

tranchitella (approved):

PR #541, opened 2021-01-26
Fix deployment group being removed from object after resolving device…

alfrunes:
… ids

This was a mistake introduced by 5dc6edebb2cdb38c531ad9d4d4d09eadf9c6527f
where the group is cleared from the document after resolving the device
ids.
changelog: none


kjaskiewiczz (approved):
looks good

PR #547, opened 2021-02-04
introduce configuration deployments

kjaskiewiczz:
Issues: https://tracker.mender.io/browse/MEN-4426, https://tracker.mender.io/browse/MEN-4431

### DIFF
--- a/api/http/api_deployments.go
+++ b/api/http/api_deployments.go
@@ -638,6 +638,78 @@ func (d *DeploymentsApiHandlers) DeployToGroup(w rest.ResponseWriter, r *rest.Re
 	d.createDeployment(w, r, ctx, l, group)
 }
 
+// parseDeviceConfigurationDeploymentPathParams parses expected params
+// and check if the params are not empty
+func parseDeviceConfigurationDeploymentPathParams(r *rest.Request) (string, string, string, error) {
+	tenantID := r.PathParam("tenant")
+	if tenantID == "" {
+		return "", "", "", errors.New("tenant ID missing")
+	}
+	deviceID := r.PathParam("id")
+	if deviceID == "" {
+		return "", "", "", errors.New("device ID missing")
+	}
+	deploymentID := r.PathParam("deployment_id")
+	if deploymentID == "" {
+		return "", "", "", errors.New("deployment ID missing")
+	}
+	return tenantID, deviceID, deploymentID, nil
+}
+
+// getConfigurationDeploymentConstructorFromBody extracts configuration
+// deployment constructor from the request body and validates it
+func getConfigurationDeploymentConstructorFromBody(r *rest.Request) (
+	*model.ConfigurationDeploymentConstructor, error) {
+
+	var constructor *model.ConfigurationDeploymentConstructor
+
+	if err := r.DecodeJsonPayload(&constructor); err != nil {
+		return nil, err
+	}
+
+	if err := constructor.Validate(); err != nil {
+		return nil, err
+	}
+
+	return constructor, nil
+}
+
+// device configuration deployment handler
+func (d *DeploymentsApiHandlers) PostDeviceConfigurationDeployment(w rest.ResponseWriter, r *rest.Request) {
+	l := requestlog.GetRequestLogger(r)
+
+	// get path params
+	tenantID, deviceID, deploymentID, err := parseDeviceConfigurationDeploymentPathParams(r)
+	if err != nil {
+		rest_utils.RestErrWithLog(w, r, l, err, http.StatusBadRequest)
+		return
+	}
+
+	// add tenant id to the context
+	ctx := identity.WithContext(r.Context(), &identity.Identity{Tenant: tenantID})
+
+	constructor, err := getConfigurationDeploymentConstructorFromBody(r)
+	if err != nil {
+		d.view.RenderError(w, r, errors.Wrap(err, "Validating request body"), http.StatusBadRequest, l)
+		return
+	}
+
+	id, err := d.app.CreateDeviceConfigurationDeployment(ctx, constructor, deviceID, deploymentID)
+	switch err {
+	default:
+		d.view.RenderInternalError(w, r, err, l)
+	case nil:
+		r.URL.Path = "./deployments"
+		d.view.RenderSuccessPost(w, r, id)
+	case app.ErrDuplicateDeployment:
+		d.view.RenderError(w, r, err, http.StatusConflict, l)
+	case app.ErrInvalidDeploymentID:
+		d.view.RenderError(w, r, err, http.StatusBadRequest, l)
+	}
+
+	return
+}
+
 func (d *DeploymentsApiHandlers) getDeploymentConstructorFromBody(r *rest.Request, group string) (*model.DeploymentConstructor, error) {
 	var constructor *model.DeploymentConstructor
 	if err := r.DecodeJsonPayload(&constructor); err != nil {
--- a/api/http/api_deployments.go
+++ b/api/http/api_deployments.go
@@ -645,7 +645,7 @@ func parseDeviceConfigurationDeploymentPathParams(r *rest.Request) (string, stri
 	if tenantID == "" {
 		return "", "", "", errors.New("tenant ID missing")
 	}
-	deviceID := r.PathParam("id")
+	deviceID := r.PathParam("device_id")
 	if deviceID == "" {
 		return "", "", "", errors.New("device ID missing")
 	}
--- a/api/http/api_deployments.go
+++ b/api/http/api_deployments.go
@@ -1045,6 +1045,18 @@ func ParseLookupQuery(vals url.Values) (model.Query, error) {
 
 	}
 
+	dType := vals.Get("type")
+	if dType == "" {
+		return query, nil
+	}
+	deploymentType := model.DeploymentType(dType)
+	if deploymentType == model.DeploymentTypeSoftware ||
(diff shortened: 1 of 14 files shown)


alfrunes (approved):
LGTM

I have a couple of minor suggestions below.

api/http/api_deployments.go
```diff
@@ -973,6 +1043,18 @@ func ParseLookupQuery(vals url.Values) (model.Query, error) {
 
 	}
 
+	dType := vals.Get("type")
+	if dType == "" {
+		return query, nil
```
tranchitella, line 1048: Shouldn't we default to `query.Type = model.DeploymentTypeSoftware` here?
Not sure if it is used somewhere else, but to me it looks cleaner to set the type in this case.

api/http/routing.go
```diff
@@ -58,11 +58,12 @@ const (
...
-	ApiUrlInternalHealth            = ApiUrlInternal + "/health"
-	ApiUrlInternalTenants           = ApiUrlInternal + "/tenants"
-	ApiUrlInternalTenantDeployments = ApiUrlInternal + "/tenants/:tenant/deployments"
-	ApiUrlInternalTenantArtifacts   = ApiUrlInternal + "/tenants/:tenant/artifacts"
+	ApiUrlInternalAlive                          = ApiUrlInternal + "/alive"
+	ApiUrlInternalHealth                         = ApiUrlInternal + "/health"
+	ApiUrlInternalTenants                        = ApiUrlInternal + "/tenants"
+	ApiUrlInternalTenantDeployments              = ApiUrlInternal + "/tenants/:tenant/deployments"
+	ApiUrlInternalTenantArtifacts                = ApiUrlInternal + "/tenants/:tenant/artifacts"
+	ApiUrlInternalDeviceConfigurationDeployments = ApiUrlInternal + "/tenants/:tenant/devices/:id/configuration/deployments/:deployment_id"
```
tranchitella, line 66: Given that we have `/tenants/:tenant/deployments`, it comes natural to me to imagine the path to be:
`/tenants/:tenant/configuration/deployments/:deployment_id/devices/:device_id`. What do you think?

docs/internal_api.yml
```diff
@@ -369,3 +424,23 @@ definitions:
...
+            The deployments service will use it to generate configuration
+            artifact for the device.
+            The artifact will be generated when the device will ask
+            for an update.
+    required:
+      - name
+      - configuration
+    example:
+      name: config-1.1
+      configuration: "{foo:bar}"
```
tranchitella, line 446:
```suggestion
      configuration: "{\"foo\":\"bar\"}"
```

docs/management_api.yml
```diff
@@ -70,6 +70,19 @@ paths:
...
             - finished
             - pending
+        - name: type
+          in: query
+          description: |
+              Deployment type filter.
+              It is possible for the deployment to not have the type field.
+              In this case, the deployments service will treat it
+              as software deployment and return it with other
+              software deployments.
```
tranchitella, line 80: This is an internal implementation detail, I don't think we need to expose it here, do we?
In particular, we have to be sure that if you do not provide a type value, you are getting all deployments. If you set "software", you only retrieve software deployments, including also the ones which don't have the value set in the database.

model/configuration_deployment.go
```diff
@@ -0,0 +1,73 @@
...
+type ConfigurationDeploymentConstructor struct {
+	// Deployment name, required
+	Name string `json:"name"`
+
+	// A string containing a configuration object.
+	// The deployments service will use it to generate configuration
+	// artifact for the device.
+	// The artifact will be generated when the device will ask
+	// for an update.
+	Configuration string `json:"configuration"`
```
tranchitella, line 34: I'm pretty sure you had the Retries here in the enterprise version. Can we add it here, as well?
We can simply ignore the field in the OS version, but this simplifies my life in deviceconfig/workflows. :)

model/deployment.go
```diff
@@ -129,6 +133,18 @@ type Deployment struct {
...
+	// currently we are supporting two types of deployments:
+	// software and configuration
+	Type DeploymentType `json:"type,omitempty" bson:"type"`
+
+	// A string containing a configuration object.
+	// The deployments service will use it to generate configuration
+	// artifact for the device.
+	// The artifact will be generated when the device will ask
+	// for an update.
+	Configuration string `json:"configuration,omitempty" bson:"configuration"`
```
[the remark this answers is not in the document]
tranchitella: +1

store/mongo/datastore_mongo.go
```diff
@@ -1720,6 +1721,20 @@ func (db *DataStoreMongo) Find(ctx context.Context,
...
 
+	// build deployment by type part of the query
+	if match.Type != "" {
+		if match.Type == model.DeploymentTypeConfiguration {
+			andq = append(andq, bson.M{StorageKeyDeploymentType: match.Type})
+		} else if match.Type == model.DeploymentTypeSoftware {
+			andq = append(andq, bson.M{
+				"$or": []bson.M{
+					{StorageKeyDeploymentType: match.Type},
+					{StorageKeyDeploymentType: ""},
```
tranchitella, line 1732: Great, that's what I meant!

tranchitella (commented):
A few comments here and there. The important things are the path of the end-point and the Retries field.

model/deployment.go
```diff
@@ -34,11 +34,15 @@ var (
...
+type DeploymentType string
 
 const (
 	DeploymentStatusFinished   DeploymentStatus = "finished"
 	DeploymentStatusInProgress DeploymentStatus = "inprogress"
 	DeploymentStatusPending    DeploymentStatus = "pending"
+
+	DeploymentTypeSoftware      DeploymentType = "software"
+	DeploymentTypeConfiguration DeploymentType = "configuration"
 )
```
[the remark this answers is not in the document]
kjaskiewiczz: done

model/deployment.go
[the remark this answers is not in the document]
kjaskiewiczz: done

api/http/api_deployments.go
kjaskiewiczz: hmm, I thought that by default we'll return deployments of both types - software and configuration and we'll filter them only when the type is provided

api/http/routing.go
kjaskiewiczz: I was thinking that the deployment resource here is a subresource of the device (the device will have more than one configuration deployment), but it's the deployment, not the device deployment so you're probably right; I'll change it according to your suggestion

docs/management_api.yml
kjaskiewiczz: I agree that this is an implementation detail, but on the other hand, we will return to the users deployments with type=software, deployments with type=configuration, and deployments without type filed; I thought it may be confusing and it would be nice to document it somewhere; 
I don't have a strong opinion here so I can remove this comment if you want

model/configuration_deployment.go
kjaskiewiczz: sure; would it be enough to have it here or should we have it in the deployment object too?

model/deployment.go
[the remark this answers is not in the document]
kjaskiewiczz: or not; do you mean changing it in the deployment or both in the deployment and in the constructor?

api/http/routing.go
kjaskiewiczz: done

docs/internal_api.yml
kjaskiewiczz: done

docs/management_api.yml
kjaskiewiczz: done

model/configuration_deployment.go
kjaskiewiczz: done

api/http/api_deployments.go
tranchitella: Sorry, you are right, I misread the diff: this is the parsing of the query, thus no default should be applied here.

docs/management_api.yml
tranchitella: My idea was that we never return objects with the empty value: if we don't have a value in the db, we set "software" when we marshall to JSON. From the user's point of view, it's always either software or configuration, he will never see empty.

model/configuration_deployment.go
tranchitella: Thinking about it twice, as long as we do not return 400 bad request because we set the "retries" attribute, I'm fine if we do not add the field. I believe in this case it will be ignored when unmarshalling the request, right?

kjaskiewiczz: yes, it will, but maybe it's safer to have it there in case we'll treat additional fields as client errors at some point

tranchitella (approved):

model/deployment.go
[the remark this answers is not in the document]
alfrunes: Sorry for all the back and forth, but it looks like we will restrict the implementation to JSON only, because we will encode the map into the [meta-data](https://github.com/mendersoftware/mender-artifact/blob/master/Documentation/artifact-format-v3.md#meta-data) header of the artifact (see [this comment](https://tracker.mender.io/browse/MEN-4427?focusedCommentId=112782&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-112782)). So we might as well store it as an object (`map[string]interface{}`) in MongoDB.

alfrunes (approved):

PR #549, opened 2021-02-05
MEN-4427: Generate configuration artifacts on-the-fly on /deployments/next requests

alfrunes:
This PR implements a new endpoint consuming signed expiring links generated by the /deployments/next API handler if the deployment is a configuration type deployment.

NOTE: This PR is still work in progress as there's missing some unit test coverage.

### DIFF
--- a/api/http/api_deployments.go
+++ b/api/http/api_deployments.go
@@ -1044,9 +1044,15 @@ func (d *DeploymentsApiHandlers) GetDeploymentForDevice(w rest.ResponseWriter, r
 		return
 	} else if deployment.Type == model.DeploymentTypeConfiguration {
 		// Generate pre-signed URL
-		var hostName string
-		if hostName = r.Header.Get("X-Forwarded-Host"); hostName == "" {
-			hostName = d.config.PresignHostname
+		var hostName string = d.config.PresignHostname
+		if hostName == "" {
+			if hostName = r.Header.Get(hdrForwardedHost); hostName == "" {
+				d.view.RenderInternalError(w, r,
+					errors.New("presign.hostname not configured; "+
+						"unable to generate download link "+
+						" for configuration deployment"), l)
+				return
+			}
 		}
 		req, _ := http.NewRequest(
 			http.MethodGet,
--- a/api/http/api_deployments.go
+++ b/api/http/api_deployments.go
@@ -985,10 +985,12 @@ func (d *DeploymentsApiHandlers) AbortDeployment(w rest.ResponseWriter, r *rest.
 }
 
 func (d *DeploymentsApiHandlers) GetDeploymentForDevice(w rest.ResponseWriter, r *rest.Request) {
-	ctx := r.Context()
-	l := requestlog.GetRequestLogger(r)
-
-	idata := identity.FromContext(ctx)
+	var (
+		installed *model.InstalledDeviceDeployment
+		ctx       = r.Context()
+		l         = requestlog.GetRequestLogger(r)
+		idata     = identity.FromContext(ctx)
+	)
 	if idata == nil {
 		d.view.RenderError(w, r, ErrMissingIdentity, http.StatusBadRequest, l)
 		return
@@ -1009,9 +1011,21 @@ func (d *DeploymentsApiHandlers) GetDeploymentForDevice(w rest.ResponseWriter, r
 			r.URL.RawQuery = q.Encode()
 		}
 	}()
-	installed := &model.InstalledDeviceDeployment{
-		ArtifactName: q.Get(GetDeploymentForDeviceQueryArtifact),
-		DeviceType:   q.Get(GetDeploymentForDeviceQueryDeviceType),
+	if strings.EqualFold(r.Method, http.MethodPost) {
+		// POST
+		installed = new(model.InstalledDeviceDeployment)
+		if err := r.DecodeJsonPayload(&installed); err != nil {
+			d.view.RenderError(w, r,
+				errors.Wrap(err, "invalid schema"),
+				http.StatusBadRequest, l)
+			return
+		}
+	} else {
+		// GET or HEAD
+		installed = &model.InstalledDeviceDeployment{
+			ArtifactName: q.Get(GetDeploymentForDeviceQueryArtifact),
+			DeviceType:   q.Get(GetDeploymentForDeviceQueryDeviceType),
+		}
 	}
 
 	if err := installed.Validate(); err != nil {
@@ -1057,45 +1071,6 @@ func (d *DeploymentsApiHandlers) GetDeploymentForDevice(w rest.ResponseWriter, r
 		}
 	}
 
-	d.view.RenderSuccessGet(w, deployment)
-}
-
-func (d *DeploymentsApiHandlers) PostDeploymentForDevice(w rest.ResponseWriter, r *rest.Request) {
-	ctx := r.Context()
-	l := requestlog.GetRequestLogger(r)
-
-	idata := identity.FromContext(ctx)
-	if idata == nil {
-		d.view.RenderError(w, r, ErrMissingIdentity, http.StatusBadRequest, l)
-		return
-	}
-
-	var installed model.InstalledDeviceDeployment
-	if err := r.DecodeJsonPayload(&installed); err != nil {
-		d.view.RenderError(w, r,
-			errors.Wrap(err, "invalid schema"),
-			http.StatusBadRequest, l)
-		return
-	}
-
-	if err := installed.Validate(); err != nil {
-		d.view.RenderError(w, r, err, http.StatusBadRequest, l)
-		return
-	}
-
-	deployment, err := d.app.GetDeploymentForDeviceWithCurrent(ctx, idata.Subject, &installed)
-	if err != nil {
-		d.view.RenderInternalError(w, r, err, l)
-		return
-	}
-
-	if deployment == nil {
-		d.view.RenderNoUpdateForDevice(w)
-		return
-	}
-
-	// NOTE: Must use the RenderSuccessGet as the POST variant reports
-	//       incorrect status code.
 	d.view.RenderSuccessGet(w, deployment)
 }
 
--- a/api/http/api_deployments.go
+++ b/api/http/api_deployments.go
@@ -85,13 +85,23 @@ var (
 )
 
 type Config struct {
+	// URL signing parameters:
+
+	// PresignSecret holds the secret value used by the signature algorithm.
 	PresignSecret []byte
+	// PresignExpire duration until the link expires.
 	PresignExpire time.Duration
+	// PresignHostname fallback value for the signed url hostname.
+	PresignHostname string
+	// PresignScheme URL scheme to used for generating signed URLs.
+	PresignScheme string
 }
 
 func NewConfig() *Config {
 	return &Config{
-		PresignExpire: time.Second * 900,
+		PresignExpire:   time.Second * 900,
+		PresignScheme:   "https",
+		PresignHostname: "localhost",
 	}
 }
 
@@ -102,6 +112,16 @@ func (conf *Config) SetPresignSecret(key []byte) *Config {
 
 func (conf *Config) SetPresignExpire(duration time.Duration) *Config {
 	conf.PresignExpire = duration
+	return conf
+}
+
+func (conf *Config) SetPresignHostname(hostname string) *Config {
+	conf.PresignHostname = hostname
+	return conf
+}
+
+func (conf *Config) SetPresignScheme(scheme string) *Config {
+	conf.PresignScheme = scheme
 	return conf
 }
 
@@ -118,7 +138,7 @@ func NewDeploymentsApiHandlers(
 	app app.App,
 	config ...*Config,
 ) *DeploymentsApiHandlers {
-	var conf Config
+	conf := NewConfig()
 	for _, c := range config {
 		if c == nil {
 			continue
@@ -129,12 +149,18 @@ func NewDeploymentsApiHandlers(
 		if c.PresignExpire != 0 {
 			conf.PresignExpire = c.PresignExpire
 		}
+		if c.PresignHostname != "" {
+			conf.PresignHostname = c.PresignHostname
+		}
+		if c.PresignScheme != "" {
+			conf.PresignScheme = c.PresignScheme
+		}
 	}
 	return &DeploymentsApiHandlers{
 		store:  store,
 		view:   view,
 		app:    app,
-		config: conf,
+		config: *conf,
 	}
 }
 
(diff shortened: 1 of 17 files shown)


kjaskiewiczz (approved):
great job!
the task was very complex IMO and you managed to finish it very quickly!
I can't see issues with this implementation, but as I mentioned, it's complex so it would be great if one more person could take a look.

config/config.go
```diff
@@ -73,6 +73,33 @@ const (
...
+	// downloadable URLs. This option is currently ignored.
+	SettingPresignAlgorithm        = "presign.algorithm"
+	SettingPresignAlgorithmDefault = "HMAC256"
+
+	// SettingPresignSecret sets the secret for generating signed url.
+	// For HMAC type of algorithms the value must be a base64 encoded
+	// secret. For public key signatures, the value must be a path to
+	// the private key (not yet supported).
+	SettingPresignSecret        = "presign.secret"
+	SettingPresignSecretDefault = "qF4/MZzQmHTJ+nrlr26b3g=="
```
tranchitella, line 87: I don't think it's a good idea to provide a default here, as this can lead to security issues.
Can we find a better way to deal with it? A random value at start-up is good enough, IMHO.

alfrunes: Good call!
Actually I discovered that I incorrectly configured the defaults below as well.
I pushed new commit generating making the secret default to a random base64 string of 32 bytes if not configured explicitly.

tranchitella (approved):

api/http/api_deployments.go
```diff
@@ -236,6 +265,80 @@ func (d *DeploymentsApiHandlers) DownloadLink(w rest.ResponseWriter, r *rest.Req
...
+	if d.config.PresignSecret == nil {
+		rest.NotFound(w, r)
+		return
+	}
+
+	var (
+		tenantID     string
+		deviceID     = r.PathParam("device_id")
+		deviceType   = r.PathParam("device_type")
+		deploymentID = r.PathParam("deployment_id")
```
merlin-northern, line 315: I would really love to see those as consts.

api/http/api_deployments.go
```diff
@@ -236,6 +265,80 @@ func (d *DeploymentsApiHandlers) DownloadLink(w rest.ResponseWriter, r *rest.Req
...
+		l            = log.FromContext(r.Context())
+		q            = r.URL.Query()
+		err          error
+	)
+	tenantID = q.Get("tenant_id")
```
merlin-northern, line 320: I would really love to see that as a const/var

api/http/api_deployments.go
```diff
@@ -994,6 +1005,26 @@ func (d *DeploymentsApiHandlers) GetDeploymentForDevice(w rest.ResponseWriter, r
 	if deployment == nil {
 		d.view.RenderNoUpdateForDevice(w)
 		return
+	} else if deployment.Type == model.DeploymentTypeConfiguration {
+		// Generate pre-signed URL
+		hostName := r.Header.Get("X-Forwarded-Host")
```
merlin-northern: a const ehre, perhaps?

api/http/routing.go
```diff
@@ -102,9 +105,15 @@ func NewRouter(ctx context.Context, c config.Reader,
 
 	app := app.NewDeployments(mongoStorage, fileStorage, app.ArtifactContentType)
 
-	apiConf := NewConfig()
-	if key, err := base64.StdEncoding.DecodeString(
-		c.GetString(dconfig.SettingPresignSecretBase64),
+	// Encode base64 secret in either std or URL encoding ignoring padding.
+	base64Repl := strings.NewReplacer("-", "+", "_", "/", "=", "")
```
merlin-northern, line 111: Maybe it is late, but I do not understand this replacing, why is it so?

api/http/routing.go
```diff
@@ -222,3 +231,12 @@ func ReleasesRoutes(controller *DeploymentsApiHandlers) []*rest.Route {
 		rest.Get(ApiUrlManagementReleases, controller.GetReleases),
 	}
 }
+
+func FMTConfigURL(hostname, deploymentID, deviceType, deviceID string) string {
+	repl := strings.NewReplacer(
+		":deployment_id", deploymentID,
+		":device_type", deviceType,
+		":device_id", deviceID,
```
merlin-northern, line 246: on more reason to have those in consts.

api/http/routing.go
```diff
@@ -222,3 +231,12 @@ func ReleasesRoutes(controller *DeploymentsApiHandlers) []*rest.Route {
...
+	)
+	return fmt.Sprintf("https://%s%s", hostname, repl.Replace(ApiUrlDevicesDownloadConfig))
```
merlin-northern: Isn't Sprintf and over kill here? `"https://"+hostname+repl.Replace(ApiUrlDevicesDownloadConfig)` perhaps?

api/http/api_deployments.go
```diff
@@ -85,13 +85,23 @@ var (
...
 type Config struct {
+	// URL signing parameters:
+
+	// PresignSecret holds the secret value used by the signature algorithm.
 	PresignSecret []byte
+	// PresignExpire duration until the link expires.
 	PresignExpire time.Duration
+	// PresignHostname fallback value for the signed url hostname.
+	PresignHostname string
+	// PresignScheme URL scheme to used for generating signed URLs.
```
merlin-northern, line 96: nitpick: "to used" -> "to _be_ used" ot "to _use_"?

api/http/api_deployments.go
```diff
@@ -85,13 +85,23 @@ var (
...
+	PresignScheme string
 }
 
 func NewConfig() *Config {
 	return &Config{
-		PresignExpire: time.Second * 900,
+		PresignExpire:   time.Second * 900,
```
merlin-northern, line 102: I know, not part of the commit, but as long as you are here, maybe some explanation on the value of 900 and a const?

api/http/api_deployments.go
```diff
@@ -1007,10 +1033,17 @@ func (d *DeploymentsApiHandlers) GetDeploymentForDevice(w rest.ResponseWriter, r
 		return
 	} else if deployment.Type == model.DeploymentTypeConfiguration {
 		// Generate pre-signed URL
-		hostName := r.Header.Get("X-Forwarded-Host")
+		var hostName string
+		if hostName = r.Header.Get("X-Forwarded-Host"); hostName == "" {
```
merlin-northern, line 1051: already used second time, maybe a const? (or this commit changes the previous)

api/http/api_deployments.go
```diff
@@ -1007,10 +1033,17 @@ func (d *DeploymentsApiHandlers) GetDeploymentForDevice(w rest.ResponseWriter, r
...
+			hostName = d.config.PresignHostname
+		}
 		req, _ := http.NewRequest(
 			http.MethodGet,
-			FMTConfigURL(hostName, deployment.ID, installed.DeviceType, idata.Subject),
+			FMTConfigURL(
+				d.config.PresignScheme, hostName,
+				deployment.ID, installed.DeviceType,
+				idata.Subject,
+			),
```
merlin-northern, line 1060: and the reason behind this formatting?

api/http/routing.go
```diff
@@ -232,11 +238,11 @@ func ReleasesRoutes(controller *DeploymentsApiHandlers) []*rest.Route {
...
 
-func FMTConfigURL(hostname, deploymentID, deviceType, deviceID string) string {
+func FMTConfigURL(scheme, hostname, deploymentID, deviceType, deviceID string) string {
 	repl := strings.NewReplacer(
 		":deployment_id", deploymentID,
 		":device_type", deviceType,
 		":device_id", deviceID,
 	)
-	return fmt.Sprintf("https://%s%s", hostname, repl.Replace(ApiUrlDevicesDownloadConfig))
+	return fmt.Sprintf("%s://%s%s", scheme, hostname, repl.Replace(ApiUrlDevicesDownloadConfig))
```
merlin-northern, line 248: now I am sure the commits change the previous commits. the same comment: how about simple concat?

config/config.go
```diff
@@ -74,11 +74,32 @@ const (
...
+
+	// SettingPresignExpireSeconds sets the amount of seconds it takes for
+	// the signed URL to expire.
+	SettingPresignExpireSeconds        = "presign.expire_seconds"
 	SettingPresignExpireSecondsDefault = 900
```
merlin-northern, line 92: I am quite sure I saw 900 in the code as a magic number.

config/config.go
```diff
@@ -144,7 +165,10 @@ var (
 		{Key: SettingsAwsTagArtifact, Value: SettingsAwsTagArtifactDefault},
 		{Key: SettingInventoryAddr, Value: SettingInventoryAddrDefault},
 		{Key: SettingInventoryTimeout, Value: SettingInventoryTimeoutDefault},
-		{Key: SettingPresignSecretBase64, Value: SettingPresignSecretBase64},
+		{Key: SettingPresignAlgorithm, Value: SettingPresignAlgorithmDefault},
+		{Key: SettingPresignSecret, Value: SettingPresignSecret},
```
merlin-northern: shouldn't the second one be `SettingPresignSecretDefault` or something similar?

main.go
```diff
@@ -90,6 +91,16 @@ func doMain(args []string) {
 		config.Config.SetEnvPrefix("DEPLOYMENTS")
 		config.Config.AutomaticEnv()
 		config.Config.SetEnvKeyReplacer(strings.NewReplacer(".", "_", "-", "_"))
+		if config.Config.Get(dconfig.SettingPresignSecret) ==
+			dconfig.SettingPresignSecretDefault {
+			l.Warn("Using the default signature secret; THIS IS A POTENTIAL SECURITY ISSUE!")
+			l.Warnf(
+				"Please reconfigure the secret in '%s' or "+
+					"override with environment variable "+
+					"DEPLOYMENTS_PRESIGN_SECRET_BASE64",
```
merlin-northern: that formatting seems strange just a bit?

api/http/api_deployments_test.go
```diff
@@ -17,20 +17,24 @@ package http
...
 	"github.com/mendersoftware/deployments/app"
 	mapp "github.com/mendersoftware/deployments/app/mocks"
 	"github.com/mendersoftware/deployments/model"
 	"github.com/mendersoftware/deployments/utils/restutil/view"
 	h "github.com/mendersoftware/deployments/utils/testing"
 	"github.com/mendersoftware/go-lib-micro/requestid"
 	"github.com/mendersoftware/go-lib-micro/rest_utils"
+	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
```
merlin-northern: don't we group mendersoftware imports separately?

app/images_test.go
```diff
@@ -534,3 +538,136 @@ func TestGenerateImageSuccessfulWithTenant(t *testing.T) {
...
+			uuid.NameSpaceOID,
+			[]byte("deployment"),
+		).String(),
+		Deployment: &model.Deployment{
+			Id: uuid.NewSHA1(
+				uuid.NameSpaceOID,
+				[]byte("deployment"),
+			).String(),
+			Type:          model.DeploymentTypeConfiguration,
+			Configuration: []byte("{\"foo\":\"bar\"}"),
```
merlin-northern, line 568: at the moment we are only supporting string values, I wonder how much of it should be tested here? perhaps a test that shows that we do not even car if it is json?

app/app.go
```diff
@@ -45,8 +45,8 @@ import (
 const (
 	ArtifactContentType              = "application/vnd.mender-artifact"
 	ArtifactConfigureType            = "mender-configure"
-	ArtifactConfigureProvides        = "module-image.mender-configure.version"
-	ArtifactConfigureProvidesCleared = "module-image.mender-configure.*"
+	ArtifactConfigureProvides        = "data-partition.mender-configure.version"
+	ArtifactConfigureProvidesCleared = "data-partition.mender-configure.*"
```
merlin-northern, line 49: I would really enjoy some comments here, where that comes from, etc.

config/config.go
mchalski: +1, that was my comment from writing acceptance tests - the default wasn't picked up, and it was 'good' imo.

api/http/api_deployments.go
```diff
@@ -891,50 +1045,38 @@ func (d *DeploymentsApiHandlers) GetDeploymentForDevice(w rest.ResponseWriter, r
 	if deployment == nil {
 		d.view.RenderNoUpdateForDevice(w)
 		return
+	} else if deployment.Type == model.DeploymentTypeConfiguration {
+		// Generate pre-signed URL
+		var hostName string
+		if hostName = r.Header.Get("X-Forwarded-Host"); hostName == "" {
```
mchalski, line 1051: very good idea on having the presign host, and even the presign scheme. you could say we both honor the X-Forwarded header, and have a configuration fallback, right?

I think I would revert the logic here though. what if we promote the config setting:
- if the setting is there - use that
- if the setting is not there - use X-Forwarded-Host
- otherwise, fail

 the reason is I'm having second thoughts. I know I cofirmed that we can trust X-Forwarded-Host, because I'm pretty sure we can trust traefik to set it. I don't know however if in a live aws setup, with it's load balancer, we will get the original DNS name of the gateway here.
 
this will be researched and POCed still, so just a thought.

mchalski (approved):
great job.
I'm ready to approve, just left 1 discussion point

api/http/routing.go
alfrunes: It converts `base64.URLEncoding` to `base64.RawStdEncoding`. This way we can decode all 4 variants of base64 encoding (with or without padding).

api/http/api_deployments.go
alfrunes: Fair point, I'll change the logic to prefer the configuration setting and blank out the default config value.

api/http/api_deployments_test.go
alfrunes: I'm not sure why it wasn't formatted like that here...

main.go
alfrunes: Suggestions?

api/http/routing.go
alfrunes: Sure

api/http/api_deployments.go
alfrunes: It's just the default from the config file.

alfrunes: LOL, I had created a constant without even using it :see_no_evil:

api/http/api_deployments.go
alfrunes: Suggestions?

alfrunes:
I've addressed your comments @mchalski and @merlin-northern and rebased on top of current master as there was a small merge conflict.

merlin-northern:
> I've addressed your comments @mchalski and @merlin-northern and rebased on top of current master as there was a small merge conflict.

right! thanks!

merlin-northern (approved):

PR #552, opened 2021-02-10
always return the type of the deployment with the deployment object

kjaskiewiczz:
It's possible that the deployment object in the database does not have
the "type" field. In this case, set the type to "software" when returning
the deployment object from the API.

### DIFF
--- a/model/deployment.go
+++ b/model/deployment.go
@@ -203,10 +203,15 @@ func (d *Deployment) MarshalJSON() ([]byte, error) {
 
 	slim := struct {
 		*Alias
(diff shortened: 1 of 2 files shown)


tranchitella (approved):

kjaskiewiczz:
integration failure is a spurious one - I've run tests twice and in the first run the state script test failed and in the second the demo artifact test failed;
coverage/coveralls — Coverage decreased (-75.8%) to 0.0% also looks like some issue with infra, not related to my change
can we force merge the PR?

PR #553, opened 2021-02-11
Use X-Forwarded-Host header value if presign.host config is empty

tranchitella:
Changelog: none

### DIFF
--- a/api/http/api_deployments.go
+++ b/api/http/api_deployments.go
@@ -110,9 +110,8 @@ type Config struct {
 
 func NewConfig() *Config {
 	return &Config{
-		PresignExpire:   DefaultDownloadLinkExpire,
-		PresignScheme:   "https",
-		PresignHostname: "localhost",
+		PresignExpire: DefaultDownloadLinkExpire,
+		PresignScheme: "https",
 	}
 }
 
--- a/api/http/api_deployments.go
+++ b/api/http/api_deployments.go
@@ -375,12 +375,20 @@ func (d *DeploymentsApiHandlers) DownloadConfiguration(w rest.ResponseWriter, r 
 		}
 		return
 	}
+	artifactPayload, err := ioutil.ReadAll(artifact)
+	if err != nil {
+		l.Error(err.Error())
+		d.view.RenderInternalError(w, r, err, l)
+		return
+	}
+
 	rw := w.(http.ResponseWriter)
 	hdr := rw.Header()
 	hdr.Set("Content-Disposition", `attachment; filename="artifact.mender"`)
 	hdr.Set("Content-Type", app.ArtifactContentType)
+	hdr.Set("Content-Length", strconv.Itoa(len(artifactPayload)))
 	rw.WriteHeader(http.StatusOK)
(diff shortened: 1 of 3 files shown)


tranchitella:
Spurious failure: `tests/test_demo_artifact.py::TestDemoArtifact::test_demo_artifact`

api/http/api_deployments.go
```diff
@@ -379,8 +378,16 @@ func (d *DeploymentsApiHandlers) DownloadConfiguration(w rest.ResponseWriter, r
...
 	hdr.Set("Content-Disposition", `attachment; filename="artifact.mender"`)
 	hdr.Set("Content-Type", app.ArtifactContentType)
+	artifactPayload, err := ioutil.ReadAll(artifact)
+	if err != nil {
+		// There's not anything we can do here in terms of the response.
+		rw.WriteHeader(http.StatusOK)
+		l.Error(err.Error())
+		return
+	}
+	hdr.Set("Content-Length", strconv.Itoa(len(artifactPayload)))
```
alfrunes, line 388: If you buffer the artifact in memory, it's not too late to report an internal error while reading it.
```suggestion

	artifactPayload, err := ioutil.ReadAll(artifact)
	if err != nil {
		d.view.RenderInternalError(e, r, err, l)
		return
	}
	hdr := rw.Header()
	hdr.Set("Content-Disposition", `attachment; filename="artifact.mender"`)
	hdr.Set("Content-Type", app.ArtifactContentType)
	hdr.Set("Content-Length", strconv.Itoa(len(artifactPayload)))
```

alfrunes: On the other hand, I believe `net/http` adds the content length header if the response is written in a single write.

alfrunes (changes requested):

api/http/api_deployments.go
```diff
@@ -379,8 +379,16 @@ func (d *DeploymentsApiHandlers) DownloadConfiguration(w rest.ResponseWriter, r
 	hdr := rw.Header()
 	hdr.Set("Content-Disposition", `attachment; filename="artifact.mender"`)
 	hdr.Set("Content-Type", app.ArtifactContentType)
+	artifactPayload, err := ioutil.ReadAll(artifact)
+	if err != nil {
+		// There's not anything we can do here in terms of the response.
+		rw.WriteHeader(http.StatusOK)
```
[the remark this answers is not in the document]
tranchitella: No idea, this is what  the code was doing before...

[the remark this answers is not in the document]
tranchitella: You are right, makes no sense here. Rewritten as Internal Server Error.

api/http/api_deployments.go
tranchitella: Right, reworked. Regarding the content-length, it is not there if we don't set it explicitly...

tranchitella:
@alfrunes can you please have another look to this PR? We need to unblock the mender-configure client acceptance tests.

alfrunes (approved):
LGTM

PR #581, opened 2021-04-15
MEN-4635: Allow customization of S3 settings on a per-tenant basis

0lmi:
https://tracker.mender.io/browse/MEN-4635

### DIFF
--- a/docs/internal_api.yml
+++ b/docs/internal_api.yml
@@ -80,7 +80,60 @@ paths:
               Internal API error
           schema:
             $ref: '#/definitions/Error'
-
+  /tenants/{id}/storage/settings:
+    get:
+      operationId: Get Storage Settings
+      tags:
+        - Internal API
+      summary: Get storage setting for a given tenant
+      description: >
+        Returns an object with per tenant storage layer specific settings.
+      parameters:
+        - name: id
+          in: path
+          type: string
+          description: Tenant ID
+          required: true
+      produces:
+        - application/json
+      responses:
+        200:
+          description: Successful response with all available settings.
+          schema:
+            $ref: "#/definitions/StorageSettings"
+        500:
+          description: Internal error.
+          schema:
+            $ref: "#/responses/InternalServerError"
+    put:
+      operationId: Set Storage Settings
+      tags:
+        - Internal API
+      summary: Set storage settings for a given tenant
+      description: Set the storage layer settings for a given tenant.
+      parameters:
+        - name: id
+          in: path
+          type: string
+          description: Tenant ID
+          required: true
+        - name: settings
+          in: body
+          description: Settings to set.
+          required: true
+          schema:
+            $ref: "#/definitions/StorageSettings"
+      responses:
+        204:
+          description: Settings updated.
+        400:
+          description: The request body is malformed.
+          schema:
+            $ref: "#/definitions/Error"
+        500:
+          description: Internal server error.
+          schema:
+            $ref: "#/definitions/Error"
   /tenants/{id}/limits/storage:
     get:
       operationId: Get Storage Usage
@@ -348,6 +401,40 @@ definitions:
     example:
       error: "error message"
       request_id: "f7881e82-0492-49fb-b459-795654e7188a"
+  StorageSettings:
+    description: Per tenant storage settings.
+    type: object
+    properties:
+      region:
+        type: string
+        description: AWS region.
+      bucket:
+        type: string
+        description: Bucket name.
+      uri:
+        type: string
+        description: Endpoint URI.
+      key:
+        type: string
+        description: Access key id (for S3 - AWS_ACCESS_KEY_ID).
+      secret:
+        type: string
+        description: Secret access key (for S3 - AWS_SECRET_ACCESS_KEY).
+      token:
+        type: string
+        description: Token.
+    required:
+      - region
+      - bucket
+      - key
+      - secret
+    example:
+      region: us-east-1
+      bucket: mender-artifacts-unique-bucket-name
+      uri: example.com
+      key: AKIAQIHEV11XZJFVCNM4
+      secret: f2Ref4JU3eRqAFO23qpy5XOhcfTnzJivAcNO1lvn
+      token: <token>
   StorageUsage:
     description: Tenant account storage limit and storage usage.
     type: object
--- a/model/storage_settings.go
+++ b/model/storage_settings.go
@@ -1,0 +1,63 @@
+// Copyright 2021 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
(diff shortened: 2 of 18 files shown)


docs/internal_api.yml
```diff
@@ -80,7 +80,67 @@ paths:
...
+      parameters:
+        - name: id
+          in: path
+          type: string
+          description: Tenant ID
+          required: true
+        - name: settings
+          in: body
+          description: >
+            New document with settings. Existing document will be overwritten if exists.
```
tranchitella, line 129:
```suggestion
            Settings to set.
```

docs/internal_api.yml
```diff
@@ -80,7 +80,67 @@ paths:
...
+          schema:
+            $ref: "#/responses/InternalServerError"
+    put:
+      operationId: Set Storage Settings
+      tags:
+        - Internal API
+      summary: Set/update storage settings for a given tenant
+      description: >
+        Save per tenant storage layer settings. 
+        The settings are saved without encryption.
```
tranchitella, line 119:
```suggestion
        Set the storage layer settings for a given tenant.
```

docs/internal_api.yml
```diff
@@ -348,6 +408,48 @@ definitions:
...
+        description: Secret access key (for S3 - AWS_SECRET_ACCESS_KEY).
+      token:
+        type: string
+        description: Token.
+      download_expire_seconds:
+        type: integer
+        description: Number of second a presigned download URL is valid.
+      upload_expire_seconds:
+        type: integer
+        description: Number of second a presigned upload URL is valid.
```
tranchitella, line 438: Let's keep download_expire_seconds and upload_expire_seconds out. I don't think these will need to be customized on a per-tenant level, at least not at this stage.

docs/internal_api.yml
```diff
@@ -348,6 +408,48 @@ definitions:
...
+      - secret
+    example:
+      region: us-east-1
+      bucket: mender-artifacts-unique-bucket-name
+      uri: mybucket.s3.us-west-2.amazonaws.com
+      key: AKIAQIHEV11XZJFVCNM4
+      secret: f2Ref4JU3eRqAFO23qpy5XOhcfTnzJivAcNO1lvn
+      token: <token>
+      download_expire_seconds: 900
+      upload_expire_seconds: 3600
```
tranchitella, line 452: I would remove them here as well, obviously.

tranchitella (commented):
I'm leaving a few comments

docs/internal_api.yml
```diff
@@ -80,7 +80,67 @@ paths:
...
+  /tenants/{id}/storage/settings:
+    get:
+      operationId: Get Storage Settings
+      tags:
+        - Internal API
+      summary: Get storage setting for a given tenant
+      description: >
+        Returns an object with per tenant storage layer specific settings. 
+        Just not to forget - might be a good idea to hide in admin panel 
+        the value of AWS_SECRET_ACCESS_KEY.
```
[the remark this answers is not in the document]
0lmi: Ok. Got it.

docs/internal_api.yml
```diff
@@ -80,7 +80,67 @@ paths:
...
+          required: true
+      produces:
+        - application/json
+      responses:
+        200:
+          description: Successful response with all available settings.
+          schema:
+            $ref: "#/definitions/StorageSettings"
+        204:
+          description: Successful response with empty data set.
```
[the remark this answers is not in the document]
0lmi: Those settings won't exist for some tenants, is it better to reply 200 with empty document?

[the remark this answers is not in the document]
tranchitella: Yes, we reply with 200 and empty document.

store/datastore.go
```diff
@@ -30,6 +30,10 @@ type DataStore interface {
 	//limits
 	GetLimit(ctx context.Context, name string) (*model.Limit, error)
 
+	//storage settings
+	GetStorageSettings(ctx context.Context, tenantID string) (*model.StorageSettings, error)
```
0lmi, line 34: I'm not sure how to do it better considering porting the changes to deployments-enterprise. In `deployments` service `tenantID string` is not used but will be used in `deployments-enterprise`.
@tranchitella How to do it better - have different interfaces for those 2 services or make it unified?

tranchitella: keep tenantID here, and assume it's an empty string if multi-tenancy is not enabled

store/mongo/datastore_mongo.go
```diff
@@ -1912,3 +1920,59 @@ func (db *DataStoreMongo) ExistByArtifactId(ctx context.Context,
...
+	database := db.client.Database(mstore.DbFromContext(ctx, DatabaseName))
+	collection := database.Collection(CollectionStorageSettings)
+
+	// need to make sure that it's only one document exists in the collection
+	// query checks if a document exists in a collection and
+	// with enabled upsert it will be created or overwritten
+	filter := bson.M{
+		"_id": bson.M{
+			"$exists": true,
+		},
```
tranchitella, line 1956: Another option is using a fix value for ID (e.g. `settings`), without relying on the autogeneration of an objectID by mongo. In this case, you could use `$setOnInsert` in the update document to set the `_id` to the constant value above.

model/storage_settings.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/pkg/errors"
+)
+
+type StorageSettings struct {
+	Region string
+	Bucket string
+	Uri    string
+	Key    string
+	Secret string
+	Token  string
```
tranchitella, line 31: Here you miss the annotations, e.g. `json:"region" bson:"region"`
This is the reason why you need the capitalized keys in the acceptance tests (`Region` vs. `region`).
Can you please apply the annotations here, and use lower case keys in the acceptance tests?

tranchitella (changes requested):
I'm requesting a few changes.
Can you please also check ./store/mocks/DataStore.go? It is missing the copyright header.

model/storage_settings.go
0lmi: Got it. Thanks!

store/mongo/datastore_mongo.go
0lmi: @tranchitella you requested changes. Is the request is related to model and this is optional? or it would be good to change the query in datastore as well?

0lmi: @tranchitella not really able to get it working and based on the following error message it seems like it should not work...
```
internal error: Failed to save settings: multiple write errors: [{write errors: [{Performing an update on the path '_id' would modify the immutable field '_id'}]}, {<nil>}]
```

0lmi: finally got it working :)

tranchitella:

0lmi:
@mender-test-bot start pipeline please :)

0lmi:

tranchitella (approved):

PR #582, opened 2021-04-16
Extend deployments/ID/status endpoint to accept pause values

mchalski:
https://tracker.mender.io/browse/MEN-4570

introducing new states is simple as such. it's important however to include them in our stats/status workflows, recomputations, etc. this is the most effort here, I hope I caught every bit.

since we are slim on unit tests in important places, I also added a minor adjustment in tests to include new states.

also note that this change inadvertently bubbles up to the management api - so that's also adjusted.


mchalski:

mchalski:
wrong branch, closing

PR #594, opened 2021-05-18
MEN-3993: new deployments devices list endpoint with pagination

0lmi:
https://tracker.mender.io/browse/MEN-3993

### DIFF
--- a/docs/management_api.yml
+++ b/docs/management_api.yml
@@ -356,14 +356,16 @@ paths:
 
   /deployments/{deployment_id}/devices:
     get:
+      deprecated: true
       operationId: List Devices in Deployment
       tags:
         - Management API
       security:
         - ManagementJWT: []
       summary: |
-        Get list of all devices and their respective status for the deployment
-        with the given ID.
+        DEPRECATED: this end-point is deprecated because it doesn't support
+        pagination and will be removed in the future, please use the
+        /deployments/{deployment_id}/devices/list end-point instead.
       parameters:
         - name: deployment_id
           in: path
--- a/docs/management_api.yml
+++ b/docs/management_api.yml
@@ -386,6 +386,52 @@ paths:
         500:
           $ref: "#/responses/InternalServerError"
 
+  /deployments/{deployment_id}/devices/list:
+    get:
+      operationId: List Devices in Deployment with pagination
+      tags:
+        - Management API
+      security:
+        - ManagementJWT: []
+      summary: |
+        Get list of all devices and their respective status for the deployment
+        with the given ID.
+      parameters:
+        - name: deployment_id
+          in: path
+          description: Deployment identifier.
+          required: true
+          type: string
+        - name: page
+          in: query
+          description: Starting page.
+          required: false
+          type: number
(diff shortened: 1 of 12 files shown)


docs/management_api.yml
```diff
@@ -356,6 +356,7 @@ paths:
 
   /deployments/{deployment_id}/devices:
     get:
+      deprecated: true
       operationId: List Devices in Deployment
```
tranchitella, line 360: This is not enough to render a nice "deprecation message" in the Mender API docs website.
Please update the Summary adding a clear deprecation message, e.g.:
DEPRECATED: this end-point is deprecated because it doesn't support pagination and will be removed in the future, please use the `/deployments/{deployment_id}/devices/list` end-point instead.

docs/management_api.yml
```diff
@@ -386,6 +387,52 @@ paths:
         500:
           $ref: "#/responses/InternalServerError"
 
+  /deployments/{deployment_id}/devices/list:
+    get:
+      operationId: Devices List in Deployment
```
tranchitella, line 392: The operationId is very similar to the one we deprecated (List Devices in Deployment vs. Devices List in Deployment). As we render this in the API docs as menu item in the sidebar, I would suggest improving the name. For example, "List Devices in Deployment with pagination". We'll remove the "with pagination" as soon as we remove the deprecated EP.

store/mongo/datastore_mongo.go
```diff
@@ -264,7 +265,8 @@ var (
 	ErrDeploymentStorageCannotExecQuery   = errors.New("Cannot execute query")
 	ErrStorageInvalidInput                = errors.New("invalid input")
 
-	ErrLimitNotFound = errors.New("limit not found")
+	ErrLimitNotFound      = errors.New("limit not found")
```
tranchitella, line 268: Is it used anywhere?

tests/tests/test_deployment.py
```diff
@@ -196,6 +196,67 @@ def test_deployments_new_no_artifact(self):
...
+            for dev in devices:
+                dc = SimpleDeviceClient()
+                dc.get_next_deployment(
+                    dev.fake_token,
+                    artifact_name="different {}".format(artifact_name),
+                    device_type=dev.device_type,
+                )
+
+            # check default 'page' and 'per_page' values
+            res = self.d.client.Management_API.Devices_List_in_Deployment(
```
tranchitella, line 239: If you change the operation ID in the API docs, you'll need to adjust this method name as well

tranchitella (changes requested):
A few requests, if possible

store/mongo/datastore_mongo.go
0lmi: Actually it's not. It was also my concern, but thought maybe it's intentional fur future purposes.

0lmi: You know, actually it's in use.

tranchitella (approved):

0lmi:
@mender-test-bot start pipeline please :)

PR #603, opened 2021-06-29
MEN-4813: avoid race condition when aborting deployments

tranchitella:
If the user aborts a deployment while devices are updating their status,
it can happen a stats recalculation is performed in parallel with the
abort operation. In this case, deployment's end time is set but the
status is not "finished", making it impossible to abort the deployment.

Changelog: none

### DIFF
--- a/model/deployment.go
+++ b/model/deployment.go
@@ -16,12 +16,13 @@ package model
 
 import (
 	"encoding/json"
-	"github.com/pkg/errors"
 	"time"
 
-	"github.com/go-ozzo/ozzo-validation/v4"
+	"github.com/pkg/errors"
+
+	validation "github.com/go-ozzo/ozzo-validation/v4"
 	"github.com/go-ozzo/ozzo-validation/v4/is"
-	"github.com/satori/go.uuid"
+	uuid "github.com/satori/go.uuid"
 )
 
 // Errors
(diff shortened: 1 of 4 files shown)


tranchitella:

tranchitella:

kjaskiewiczz (approved):

model/deployment.go
```diff
@@ -16,12 +16,13 @@ package model
...
-	"github.com/pkg/errors"
 	"time"
 
-	"github.com/go-ozzo/ozzo-validation/v4"
+	"github.com/pkg/errors"
+
+	validation "github.com/go-ozzo/ozzo-validation/v4"
 	"github.com/go-ozzo/ozzo-validation/v4/is"
-	"github.com/satori/go.uuid"
+	uuid "github.com/satori/go.uuid"
```
alfrunes, line 25: Not really related to the pr, but can we change this one to github.com/google/uuid ? (ref. satori/go.uuid#115)

alfrunes (approved):

model/deployment.go
tranchitella: Yes we can, do you want me to do it now?

alfrunes: If you have time, otherwise I can do it later.

tranchitella: I will do it as a follow-up PR, then.

PR #612, opened 2021-07-21
removed deprecated ci job image override to align with ci templates

mzedel:
- only for static tests though as unit tests still need to be aligned with go:1.14

Changelog: None
Signed-off-by: Manuel Zedel <manuel.zedel@northern.tech>

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -30,8 +30,5 @@ test:acceptance_tests:
     DOCKER_HOST: tcp://docker:2375/
     DOCKER_DRIVER: overlay2
 
-test:static:
-  image: golang:1.14
-
 test:unit:
   image: golang:1.14


.gitlab-ci.yml
```diff
@@ -30,8 +30,5 @@ test:acceptance_tests:
     DOCKER_HOST: tcp://docker:2375/
     DOCKER_DRIVER: overlay2
 
-test:static:
-  image: golang:1.14
-
 test:unit:
   image: golang:1.14
```
lluiscampos, line 34: Can we also remove this one? I did so for `mender-cli`, see https://github.com/mendersoftware/mender-cli/pull/137

mzedel: no, as described in the commit message:
> - only for static tests though as unit tests still need to be aligned with go:1.14

this is due to the certificate in mendersoftware/integration#1253 - to be correct, even the static check change should not be adjusted - but I think it is tolerable... for the unit tests I don't think it is

lluiscampos: Sorry I didn't read that :see_no_evil:

lluiscampos (approved):

PR #627, opened 2021-08-31
MEN-4525: use internal end-point to decommission devices

tranchitella:
We move the decommissioning end-point from management to internal, where
it really belongs to. This commit breaks back-compatibility, but the
endpoint wasn't meant to be used by users and it is safer to remove it
straight away.

Changelog: title

### DIFF
--- a/api/http/routing.go
+++ b/api/http/routing.go
@@ -51,7 +51,6 @@ const (
 	ApiUrlManagementDeploymentsDevices     = ApiUrlManagement + "/deployments/:id/devices"
 	ApiUrlManagementDeploymentsDevicesList = ApiUrlManagement + "/deployments/:id/devices/list"
 	ApiUrlManagementDeploymentsLog         = ApiUrlManagement + "/deployments/:id/devices/:devid/log"
-	ApiUrlManagementDeploymentsDeviceId    = ApiUrlManagement + "/deployments/devices/:id"
 	ApiUrlManagementDeploymentsDeviceList  = ApiUrlManagement + "/deployments/:id/device_list"
 
 	ApiUrlManagementReleases = ApiUrlManagement + "/deployments/releases"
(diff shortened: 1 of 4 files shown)


tranchitella:

tranchitella:
Combined QA: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/362507133

tests/acceptance.f2Qkb1
```diff
@@ -0,0 +1,911 @@
+Attaching to acceptance-tests_acceptance_1, acceptance-tests_mender-deployments_1, acceptance-tests_mender-api-gateway_1, acceptance-tests_mender-device-auth_1, acceptance-tests_mender-useradm_1, acceptance-tests_mender-inventory_1, acceptance-tests_mender-workflows-server_1, acceptance-tests_mender-gui_1, acceptance-tests_mender-mongo_1, acceptance-tests_minio_1
```
alfrunes, line 1: I thought we had gitignore rules for these files...

Could you please remove these output files as well as the `tests/*_api.yml` and `tests/run-test-environment` files

api/http/routing.go
```diff
@@ -192,8 +192,6 @@ func NewDeploymentsResourceRoutes(controller *DeploymentsApiHandlers) []*rest.Ro
 			controller.GetDevicesListForDeployment),
 		rest.Get(ApiUrlManagementDeploymentsLog,
 			controller.GetDeploymentLogForDevice),
-		rest.Delete(ApiUrlManagementDeploymentsDeviceId,
```
alfrunes, line 195: Can we really remove this endpoint with no prior deprecation warning?

alfrunes (changes requested):
Please clean up the acceptance tests files ;)

tests/acceptance.f2Qkb1
tranchitella: Oh, sorry, bad mistake. I force-pushed to remove them.

api/http/routing.go
tranchitella: I personally believe we should remove it, because it is dangerous to keep it there (just like the inventory one).
Technically speaking, we should deprecate it, you are right, but I highly doubt somebody can use that end-point in the wild as it is designed to be called from the decommissioning workflow.

alfrunes (approved):

PR #653, opened 2021-12-07
Fix bucket initialization

emrodre:
Object storage platforms based on Ceph (https://ceph.io), like DigitalOcean Spaces, have an issue (https://tracker.ceph.com/issues/17398) where the function CreateBucket() does not return a ErrCodeBucketAlreadyOwnedByYou code.

This causes the following behaviour:
1. If the bucket does not exists it is created successfully.
2. If the bucket already exists it returns a 409 BucketAlreadyExists response and crashes so it's impossible to run the service.

The function HeadBucket() returns successfully if the bucket exists and the user has permissions so it can be used to check if the bucket exists and do an early return to avoid calling CreateBucket() when it already exists.

Also, there is a second possible corner case with calling CreateBucket() without checking if the bucket exists with HeadBucket() first as stated in the AWS SDK docs (https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.40.37/service/s3?utm_source=gopls#S3.CreateBucket):

> * ErrCodeBucketAlreadyOwnedByYou "BucketAlreadyOwnedByYou"
> The bucket you tried to create already exists, and you own it. Amazon S3
> returns this error in all Amazon Web Services Regions except in the North
> Virginia Region. For legacy compatibility, if you re-create an existing bucket
> that you already own in the North Virginia Region, Amazon S3 returns 200
> OK and resets the bucket access control lists (ACLs).

### DIFF
--- a/s3/filestorage.go
+++ b/s3/filestorage.go
@@ -164,6 +164,19 @@ func getArtifactByTenant(ctx context.Context, objectID string) string {
 }
 
 func (s *SimpleStorageService) InitBucket(ctx context.Context, bucket string) error {
+	hparams := &s3.HeadBucketInput{
+		Bucket: aws.String(bucket),
+	}
+
+	headBucketCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+
+	_, err := s.client.HeadBucketWithContext(headBucketCtx, hparams)
+	if nil == err {
+		// bucket exists and have permission to access it
+		return nil
+	}
+
 	// minio requires explicit bucket creation
 	cparams := &s3.CreateBucketInput{
 		Bucket: aws.String(bucket), // Required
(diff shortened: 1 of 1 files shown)


emrodre:
Link to a forum post where other users have this issue: https://hub.mender.io/t/mender-deployments-and-s3-bucket/796

tranchitella:
Thanks for your contribution @emrodre, it looks good.
Let me run the full integration tests on your PR: please, @mender-test-bot start pipeline.

tranchitella (approved):

emrodre:
Hi @tranchitella,

Will this fix be cherry picked to 3.1.x release? We are currently using our own Docker image with it included.

tranchitella:
@emrodre yes, see https://github.com/mendersoftware/deployments/pull/659

lluiscampos:
Hi @emrodre,

I am preparing Mender 3.1.1 release (and others) and came up with your contribution. For each release, we generate [these kind of statistics](https://docs.mender.io/release-information/release-notes-changelog/mender-server#statistics). Now, your e-mail domain "sentinel.tech" brings me to https://sentinel.tech/ , should I just set `SENTINEL` as your employer or is this a personal contribution?

Thanks

emrodre:
Hi @lluiscampos,

If you can set my employer as `Sentinel.tech` I would prefer it over just `SENTINEL`.

Thank you

PR #654, opened 2021-12-10
MEN-5259: add back /deployments/devices/{id} end-point to management

tranchitella:
Commit `7384a418c5a5fae3d7a71b13fb6beacef289a4e8` moved the end-point
mentioned above to internal, where it really belongs to because the
decommissioning workflow uses it. Still, the end-point makes sense. It
now aborts all the active and pending deployments for a device.

This commit brings back the old end-point, changing its behavior to mark
active and pending device deployments as aborted.

Changelog: title


tranchitella:

kjaskiewiczz (approved):
looks good

PR #655, opened 2021-12-13
model: Configurations are decoded "raw" if not a string

alfrunes:
TL;DR: The new workflow uses a JSON object for the configuration parameter which broke with the old behavior which expects an escaped string with the configuration object.

The new behavior is backward compatible in that configuration still
escapes JSON strings.

mendersoftware/azure-iot-manager#16


tranchitella (approved):
Looks good!

PR #660, opened 2021-12-19
Change alive and healthcheck status code to 200

gigilibala:
Currently these two APIs return status code 204 (no content), which is
fine but it doesn't play well with systems that require status code 200
for healthcheck and alive APIs. One of these is Kubernetes Ingress. The
Ingress by default (and cannot be changed) expects the status for
HTTP(s) healthcheck API be 200. This commit allows the deployment
service be configured behind an Ingress.

Changelog: Commit

Signed-off-by: Amin Hassani <ahassani@nuro.ai>

### DIFF
--- a/api/http/api_deployments.go
+++ b/api/http/api_deployments.go
@@ -192,7 +192,7 @@ func NewDeploymentsApiHandlers(
 }
 
 func (u *DeploymentsApiHandlers) AliveHandler(w rest.ResponseWriter, r *rest.Request) {
-	w.WriteHeader(http.StatusNoContent)
+	w.WriteHeader(http.StatusOK)
 }
 
 func (u *DeploymentsApiHandlers) HealthHandler(w rest.ResponseWriter, r *rest.Request) {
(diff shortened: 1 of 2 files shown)


gigilibala:
I know this maybe not acceptable as this changes the API. If you don't think this change is acceptable, I can also create two new version of these two APIs that return 200 status code. WDYT?

tranchitella:
@gigilibala why is your ingress performing health checks on the deployments service?
Wouldn't it be better to perform health checks on the API gateway (Traefik) instead?

gigilibala:
> 
Sorry, I was away on holidays.
I have looked for that in the past, but I was not able to find anything other than traefik's internal healthchecks for services it is routing to. But did a quick search again and found actually there is a `/ping` API that needs to be enabled. It seems like mender very recently enabled this ping here: https://github.com/mendersoftware/mender-helm/commit/6e3f4da55416f68b14e4cd2039a5230c1943c06b So, we probably can take similar approaches. I'm going to look at this, but hopefully don't get into problem with ingress again as port selection of ingress for health checks is always a trouble

tranchitella:
@gigilibala Yes, I suggest using `/healthz` which we recently exposed from Traefik using the `ping` internal service.
Let me know if you need assistance.

gigilibala:
Closing this as I was able to enable healthchecks without changing this API. Thanks.

PR #678, opened 2022-02-20
feat: store a new `active` property for deployments in mongodb (MEN-5443)

tranchitella:
Changelog: none
Ticket: MEN-5443


tranchitella:

tranchitella:
I force-pushed updating the commit message.

model/deployment.go
```diff
@@ -139,6 +140,9 @@ type Deployment struct {
 	// Status is the overall deployment status
 	Status DeploymentStatus `json:"status" bson:"status"`
 
+	// Active is true for unfinished deployments
+	Active bool `json:"-" bson:"active"`
```
mzedel, line 144: just out of curiosity: is there a benefit of using this over relying on the absence of the [finished timestamp](https://github.com/mendersoftware/deployments/blob/master/model/deployment.go#L126)?

tranchitella: We need a partial index (filter by X, sort by Y). The documentation is here:
https://docs.mongodb.com/manual/core/index-partial/#create-a-partial-index

Unfortunately, "$not" is not supported. It means you can create a partial index for "$exists", but not for the negation of this clause. In our case, we want to get the active deployments (pending/inprogress). Therefore, we need the `{"$not": { "$exists": true}}`.

mzedel: Good to know, thank you!

store/mongo/migration_1_2_8_test.go
```diff
@@ -0,0 +1,95 @@
...
+				IndexDeploymentsActiveCreated,
+			}
+			// verify new indexes are present
+			collection := c.Database(tc.db).Collection(CollectionDeployments)
+			indexes := collection.Indexes()
+			cursor, _ := indexes.List(ctx)
+			for cursor.Next(ctx) {
+				var tmp map[string]interface{}
+				_ = cursor.Decode(&tmp)
+				t.Log(tmp)
```
alfrunes, line 86: Should there be any assertions here?

alfrunes (approved):

store/mongo/migration_1_2_8_test.go
tranchitella: Uhm, don't know. I think this is there for debugging purposes. Actually it's a copy-and-paste from `migration_1_2_7_test.go`. It simply logs all the indexes so in case it fails, you have the list of indexes in the db.

PR #682, opened 2022-03-04
fix: change the error handling when the device is reporting wrong artifact name

kjaskiewiczz:
In case the device is reporting new artifact name before finishing the
update process, log the error and continue the update.


kjaskiewiczz:

tranchitella (approved):

MaciejTe (approved):

PR #687, opened 2022-03-10
fix the device_type handling in the deployments/next handler

kjaskiewiczz:
On the device_type change, deployments service is trying to select new
artfiact for the device deployment.
With this change, deployments service will:
- save the device_type with the device deployment object;
- try to update the artifact assigned to the device deployment
  only if the device_type has changed and the device deployment
  is not in progress

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1136,12 +1136,16 @@ func (d *Deployments) assignArtifact(
 	}
 
 	if err := d.db.AssignArtifact(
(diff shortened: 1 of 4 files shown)


kjaskiewiczz:

tranchitella:
Failure in `test_deployment_two_devices_parallel_updates_one_failure` is not relevant.

tranchitella (approved):

PR #691, opened 2022-03-17
perf: Use indexes to sort queries for next and last deployment for device

alfrunes:
…vice

This commit avoids performing in memory sort, and simplifies the query
slightly utilising the enumerated deployment status ranges.
However, this still perform an index scan which is suboptimal, but it is
as good as it gets using the current data model and indexes.

Changelog: none
Ticket: MEN-5521

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1156,10 +1156,7 @@ func (d *Deployments) getDeploymentForDevice(ctx context.Context,
 	deviceID string) (*model.Deployment, *model.DeviceDeployment, error) {
 
 	// Retrieve device deployment
-	deviceDeployment, err := d.db.FindOldestDeploymentForDeviceIDWithStatuses(
-		ctx,
-		deviceID,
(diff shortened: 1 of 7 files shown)


tranchitella:
@alfrunes I'm not convinced; I see your point (no need to create yet another index). On the other hand, these are queries we run for each `deployments/next` API call. I would rather optimize for the reading than writing. The specific index will penalize the write operations, which are rare compared to the read operations. What do you think?

alfrunes:
> @alfrunes I'm not convinced; I see your point (no need to create yet another index). On the other hand, these are queries we run for each `deployments/next` API call. I would rather optimize for the reading than writing. The specific index will penalize the write operations, which are rare compared to the read operations. A partial index would be extremely efficient, here. What do you think?

I was actually about to discard this approach when I came to the same realization, but I wanted to get your feedback first. I'm closing this one and prepare a schema migration instead.

PR #698, opened 2022-03-22
fix: improve error handling in the /deployments/next handlers

kjaskiewiczz:
Issue: https://tracker.mender.io/browse/MEN-5536

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"io"
 	"io/ioutil"
+	"reflect"
 	"strings"
 	"time"
 
@@ -92,6 +93,7 @@ var (
 	ErrNoDevices               = errors.New("No devices for the deployment")
 	ErrDuplicateDeployment     = errors.New("Deployment with given ID already exists")
 	ErrInvalidDeploymentID     = errors.New("Deployment ID must be a valid UUID")
+	ErrConflictingRequestData  = errors.New("Device provided conflicting request data")
 )
 
 //deployments
@@ -1140,7 +1142,7 @@ func (d *Deployments) assignArtifact(
 		deviceDeployment.DeviceId,
 		deviceDeployment.DeploymentId,
 		artifact,
-		installed.DeviceType,
+		installed,
 	); err != nil {
 		return errors.Wrap(err, "Assigning artifact to the device deployment")
 	}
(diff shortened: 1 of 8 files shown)


tranchitella:
@kjaskiewiczz failures in `tests/test_db_migration.py::TestDBMigrationOpenSource::test_migrate_from_legacy_mender_v1_failure` are bad: it means we are breaking the upgrade path from Mender client version 1 (no database) to version 2 (with database). What shall we do here? @kacf opinions?

app/app.go
```diff
@@ -1294,32 +1294,49 @@ func (d *Deployments) GetDeploymentForDeviceWithCurrent(ctx context.Context, dev
...
-	if installed.ArtifactName != "" && deployment.ArtifactName == installed.ArtifactName {
-		// if the device reported same artifact name as the one in the device deployment,
-		// and this is a new device deployment - indicated by device deployment
-		// status "pending",
-		// pretend there is no deployment for this device, but update
-		// its status to already installed first
-		if deviceDeployment.Status == model.DeviceDeploymentStatusPending {
+	if deviceDeployment.Request != nil &&
+		(deviceDeployment.Request.DeviceType != installed.DeviceType ||
+			deviceDeployment.Request.ArtifactName != installed.ArtifactName) {
```
alfrunes, line 1299: nit: This can be simplified a little:
```suggestion
	if deviceDeployment.Request != nil && !reflect.DeepEqual(deviceDeployment.Request, installed) {
```

app/app.go
```diff
@@ -1294,32 +1294,49 @@ func (d *Deployments) GetDeploymentForDeviceWithCurrent(ctx context.Context, dev
...
+		if err := d.UpdateDeviceDeploymentStatus(ctx, deviceDeployment.DeploymentId, deviceID,
+			model.DeviceDeploymentState{
+				Status: model.DeviceDeploymentStatusFailure,
+			}); err != nil {
 
-				return nil, errors.Wrap(err, "Failed to update deployment status")
-			}
+			return nil, errors.Wrap(err, "Failed to update deployment status")
+		}
+		return nil, nil
```
alfrunes, line 1317: Should there be an error definition for this one?
I think this deserves a HTTP/409 Conflict response - this will also make the client abort the deployment.

kjaskiewiczz: I'm not sure what client will do exactly; I guess it will at least retry in case of error, right? there is no special handling for 409 error in the client

kjaskiewiczz:

app/app.go
alfrunes: I believe it will fail the deployment if it receives an unexpected status code.

app/app.go
kjaskiewiczz: as you wish!

app/app.go
kjaskiewiczz: done

kjaskiewiczz:

kjaskiewiczz:
test failure looks unrelated - TestConfiguration.test_configuration

store/mongo/datastore_mongo.go
```diff
@@ -368,7 +368,7 @@ const (
 	StorageKeyDeviceDeploymentFinished       = "finished"
 	StorageKeyDeviceDeploymentIsLogAvailable = "log"
 	StorageKeyDeviceDeploymentArtifact       = "image"
-	StorageKeyDeviceDeploymentDeviceType     = "devicetype"
+	StorageKeyDeviceDeploymentRequest        = "request"
```
tranchitella, line 371: What will happen to the device deployments in progress when we upgrade the deployments service? From what I read, it should just skip the check and proceed as before. Can you please confirm?

alfrunes (approved):

PR #720, opened 2022-05-09
MEN-5602 and MEN-5599

tranchitella:
fix: return 404 instead of 500 when a device updates non-existent deployments
fix(docs): DeploymentInstructions's id property is not device unique

Depends on: https://github.com/mendersoftware/integration/pull/2015


tranchitella:

kjaskiewiczz (approved):

alfrunes (approved):

PR #750, opened 2022-07-07
fix: Escape configuration download URL path and support '.' in parameter

alfrunes:
Changelog: Title
Ticket: MEN-5712

### DIFF
--- a/api/http/routing.go
+++ b/api/http/routing.go
@@ -17,6 +17,7 @@ package http
 import (
 	"context"
 	"encoding/base64"
+	"net/url"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


alfrunes:

api/http/routing.go
```diff
@@ -42,45 +43,45 @@ const (
 	ApiUrlManagementArtifacts           = ApiUrlManagement + "/artifacts"
 	ApiUrlManagementArtifactsList       = ApiUrlManagement + "/artifacts/list"
 	ApiUrlManagementArtifactsGenerate   = ApiUrlManagement + "/artifacts/generate"
-	ApiUrlManagementArtifactsId         = ApiUrlManagement + "/artifacts/:id"
-	ApiUrlManagementArtifactsIdDownload = ApiUrlManagement + "/artifacts/:id/download"
+	ApiUrlManagementArtifactsId         = ApiUrlManagement + "/artifacts/#id"
```
tranchitella, line 46: Do we really need to move all placeholder prefix from the colon to the sharp?

alfrunes: For the [go-json-rest](https://pkg.go.dev/github.com/ant0ine/go-json-rest/rest#Route) package, using colon for path parameters means that it will parse up to the next `/` or `.` in the matching string, whereas the sharp `#` will treat the `.` as part of the path parameter which is what we want in all cases anyway.

tranchitella: oh, man, I didn't know it

kjaskiewiczz (approved):

kjaskiewiczz:
@alfrunes @tranchitella should we have a follow up task to change `:` to `#` in other services which are using go-json-rest?

PR #762, opened 2022-08-11
feat(bootstrap-artifact): Support for empty Artifacts

lluiscampos:
Amends commit 4f879a551de08cf1d7896de521950aaafa81ad30

Modify ArtifactUpdateTypeInfo struct and corresponding API spec to
accept either nil or string for type.

Implemented acceptance tests to cover basic operations with bootstrap
Artifacts.

Ticket: MEN-2583
Changelog: ArtifactTypeInfo object returned in `/artifacts` or
`/deployments/releases` can now return `nil` as its type. This would be
the case for bootstrap Artifacts.

### DIFF
--- a/model/update.go
+++ b/model/update.go
@@ -1,4 +1,4 @@
-// Copyright 2019 Northern.tech AS
+// Copyright 2022 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
@@ -18,7 +18,7 @@ import "time"
 
 // Type info structure
 type ArtifactUpdateTypeInfo struct {
-	Type string `json:"type" valid:"required"`
+	Type *string `json:"type" valid:"required"`
 }
 
 // Update file structure
--- a/tests/common.py
+++ b/tests/common.py
@@ -1,5 +1,5 @@
 #!/usr/bin/python
-# Copyright 2021 Northern.tech AS
+# Copyright 2022 Northern.tech AS
 #
 #    Licensed under the Apache License, Version 2.0 (the "License");
 #    you may not use this file except in compliance with the License.
@@ -23,6 +23,8 @@ import string
 import json
 import pytest
 import minio
+
+from typing import List
 
 from hashlib import sha256
 from contextlib import contextmanager
@@ -113,7 +115,9 @@ def artifact_from_raw_data(data):
 
 
 @contextmanager
-def artifact_from_data(name="foo", data=None, devicetype="hammer"):
+def artifact_rootfs_from_data(
+    name: str = "foo", data: bytes = None, devicetype: str = "hammer"
+):
     with tempfile.NamedTemporaryFile(prefix="menderout") as tmender:
         logging.info("writing mender artifact to temp file %s", tmender.name)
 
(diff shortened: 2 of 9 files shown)


tests/tests/test_artifact.py
```diff
@@ -113,8 +114,8 @@ def test_artifacts_valid(self):
             update = res.updates[0]
             assert len(update.files) == 1
             uf = update.files[0]
-            assert uf.size == len(data)
-            assert uf.checksum
+            assert uf["size"] == len(data)
+            assert uf["checksum"]
```
lluiscampos, line 118: The fact that I had to modify these in the existing tests make me suspect that the changes I made in the API spec are wrong. Please look at it carefully.

lluiscampos:
@mender-test-bot start pipeline slow :snail:

tests/common.py
```diff
@@ -113,7 +113,7 @@ def artifact_from_raw_data(data):
 
 
 @contextmanager
-def artifact_from_data(name="foo", data=None, devicetype="hammer"):
+def artifact_rootfs_from_data(name="foo", data=None, devicetype="hammer"):
```
MaciejTe, line 116: nit: please add typing info :)

tests/common.py
```diff
@@ -140,6 +140,35 @@ def artifact_from_data(name="foo", data=None, devicetype="hammer"):
                 yield fa
 
 
+@contextmanager
+def artifact_bootstrap_from_data(
+    name="foo", devicetype="hammer", provides=[], clears_provides=[]
```
MaciejTe, line 145: nit: please add typing info :)

tests/common.py
```diff
@@ -140,6 +140,35 @@ def artifact_from_data(name="foo", data=None, devicetype="hammer"):
...
+):
+    with tempfile.NamedTemporaryFile(prefix="menderout") as tmender:
+        logging.info("writing mender artifact to temp file %s", tmender.name)
+
+        cmd = (
+            'mender-artifact write bootstrap-artifact --device-type "{}" '
+            '--artifact-name "{}" --output-path "{}" {} {}'.format(
```
MaciejTe, line 152: nit: please use modern python f-strings instead of .format

MaciejTe: update: also below .format is used

tests/tests/test_artifact.py
```diff
@@ -159,6 +160,80 @@ def test_artifacts_valid(self):
             else:
                 raise AssertionError("expected to fail")
 
+    @pytest.mark.usefixtures("clean_minio", "clean_db")
```
MaciejTe, line 163: Nice!

tests/tests/test_artifact.py
```diff
@@ -159,6 +160,80 @@ def test_artifacts_valid(self):
...
+    def test_artifacts_bootstrap_valid(self):
+        artifact_name = str(uuid4())
+        description = "description for foo " + artifact_name
```
MaciejTe, line 166: nit: f-string might be used here

MaciejTe (approved):

tests/tests/test_artifact.py
alfrunes: @lluiscampos The keyword `oneOf` is not in the Swagger 2.0 specification. I don't think Swagger is rich enough to express the desired API.

lluiscampos: As discussed privately, I will revert the API spec change and workaround it in the tests.

lluiscampos: Done.

lluiscampos:
@mender-test-bot start pipeline --fast :fast_forward:

model/update.go
```diff
@@ -18,7 +18,7 @@ import "time"
 
 // Type info structure
 type ArtifactUpdateTypeInfo struct {
-	Type string `json:"type" valid:"required"`
+	Type *string `json:"type" valid:"required"`
```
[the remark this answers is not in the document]
lluiscampos: I experimented a bit with this. It is "required" in the sense that it will always be there when parsing the Mender Artifact. Is only that its value is `nil`. That is why I ended up leaving it.

PR #763, opened 2022-08-18
feat: omit configuration values in the deployment details

tranchitella:
Configuration settings can contain sensitive values. Therefore, we mask
the values to limit the reask of leaking secrets through the deployments
APIs.

Changelog: Title
Ticket: None

### DIFF
--- a/model/configuration_deployment.go
+++ b/model/configuration_deployment.go
@@ -1,4 +1,4 @@
-// Copyright 2021 Northern.tech AS
+// Copyright 2022 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
@@ -21,6 +21,11 @@ import (
 
 	validation "github.com/go-ozzo/ozzo-validation/v4"
 	"github.com/pkg/errors"
+)
+
+const (
+	omitted        = "...<omitted>"
+	lengthOmission = 3
 )
 
 // configuration saves the configuration as a binary blob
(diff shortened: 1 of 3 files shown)


tranchitella:

mzedel:
@tranchitella just a heads up: this will likely break the configuration retry functionality in the UI & maybe displaying deployment info for configuration deployments...

tranchitella:
> @tranchitella just a heads up: this will likely break the configuration retry functionality in the UI & maybe displaying deployment info for configuration deployments...

Displaying is not a problem (I tested it, it correctly shows the truncated values).
For the retry, I need to test it.

tranchitella:
> > @tranchitella just a heads up: this will likely break the configuration retry functionality in the UI & maybe displaying deployment info for configuration deployments...
> 
> Displaying is not a problem (I tested it, it correctly shows the truncated values). For the retry, I need to test it.

I was thinking about it, the UI should not allow users to "retry" a configuration deployment.
The API to trigger a configuration deployment is provided by deviceconfig.

mzedel:
> I was thinking about it, the UI should not allow users to "retry" a configuration deployment.
> The API to trigger a configuration deployment is provided by deviceconfig.

Not sure if the deployment information about the configuration is also what's used in the configuration section in the device details, but if I remember correctly this is the case if the configuration deployment failed, but we want to allow applying the configuration again. This would  also use deviceconfig to apply the configuration. But I may also be wrong, as it's been a while since we introduced the config addon...

kjaskiewiczz (approved):

alfrunes (approved):

0lmi (approved):

PR #778, opened 2022-10-05
MEN-5942: Allow Azure blob storage and s3 tenant storage settings to coexist

alfrunes:


docs/internal_api.yml
```diff
@@ -433,35 +433,61 @@ definitions:
...
+          Access key identifier (Azure: account name).
       secret:
         type: string
-        description: Secret access key (for S3 - AWS_SECRET_ACCESS_KEY).
+        description: >-
+          Secret access key (Azure: access key).
       token:
         type: string
-        description: Token.
+        description: AWS S3 session token.
```
kjaskiewiczz, line 468: is this one s3 only?

docs/internal_api.yml
```diff
@@ -433,35 +433,61 @@ definitions:
...
+          Bucket/container endpoint URI.
       external_uri:
         type: string
-        description: External Endpoint URI (if differ from uri).
+        description: Public Endpoint URI for presigning URLs (s3 only).
       key:
         type: string
-        description: Access key id (for S3 - AWS_ACCESS_KEY_ID).
+        description: >-
+          Access key identifier (Azure: account name).
```
kjaskiewiczz, line 461: there is a separate property called account_name, but from this description I would assume that if I'm using azure then I should provide account name as a key here

docs/internal_api.yml
```diff
@@ -433,35 +433,61 @@ definitions:
...
       secret:
         type: string
-        description: Secret access key (for S3 - AWS_SECRET_ACCESS_KEY).
+        description: >-
+          Secret access key (Azure: access key).
```
kjaskiewiczz, line 465: similar to account name

kjaskiewiczz (commented):
What about having storage/settings/s3 and storage/setting/blob and having two different schemas for them?
How does configuring of those settings look like? Can tenant user configure it by himself?

docs/internal_api.yml
alfrunes: I added an alias for it only when using Azure for convenience. Internally, they both map to the same value.

docs/internal_api.yml
alfrunes: Yes

alfrunes:
> What about having storage/settings/s3 and storage/setting/blob and having two different schemas for them? How does configuring of those settings look like? Can tenant user configure it by himself?

I'm not sure I get the first part of your question.
I tried to keep the settings as closely aligned as possible since the parameters are more or less the same, it's primarily the naming that differs. To make the API more aligned with the Azure naming, I created aliases that maps to the same internal values. The custom settings are only accessible through the internal API and therefore must be configured by the server admin.

kjaskiewiczz:
> I tried to keep the settings as closely aligned as possible since the parameters are more or less the same

exactly this is what I'm asking about - if we should keep one structure with all the parameters, because they are similar, both, for s3 and blob storage (may be error prone? just guessing) or if we should have two objects - one with s3 setting and one with blob storage setting?
I don't have a strong opinion; I'm pointing it out just becuse when I was looking at the code, I got lost for a moment which settings are for which storage

docs/internal_api.yml
kjaskiewiczz: in other places you have this s3 only / azure only notes

kjaskiewiczz (approved):

PR #779, opened 2022-10-12
fix: Internal storage API not following PUT semantics

alfrunes:
This commit forces the tenant settings PUT endpoint to replace the entire document, and it also fixes it so that you can use the API to store azure configurations.

Changelog: Title
Ticket: None


alfrunes:

kjaskiewiczz (approved):

PR #795, opened 2022-12-03
make it possible to query by artifact provides

kjaskiewiczz:
Ticket: [Men 6107](https://tracker.mender.io/browse/MEN-6107)

### DIFF
--- a/model/image.go
+++ b/model/image.go
@@ -37,10 +37,11 @@ const (
 )
 
 var (
+	providesKeyReplacer                   = GetProvidesKeyReplacer()
 	StorageKeyImageProvidesRootFSChecksum = "meta_artifact.provides." +
-		GetProvidesKeyReplacer().Replace("rootfs-image.checksum")
+		providesKeyReplacer.Replace("rootfs-image.checksum")
 	StorageKeyImageProvidesRootFSVersion = "meta_artifact.provides." +
-		GetProvidesKeyReplacer().Replace("rootfs-image.version")
+		providesKeyReplacer.Replace("rootfs-image.version")
 )
 
 type Provides map[string]string
@@ -288,9 +289,8 @@ func GetProvidesKeyReplacer() *strings.Replacer {
 // document.
 func (p Provides) MarshalBSONValue() (bsontype.Type, []byte, error) {
 	attrs := make(bson.M, len(p))
-	replacer := GetProvidesKeyReplacer()
 	for k, v := range p {
-		attrs[replacer.Replace(k)] = v
+		attrs[providesKeyReplacer.Replace(k)] = v
 	}
 	return bson.MarshalValue(attrs)
 }
--- a/model/image.go
+++ b/model/image.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"io"
 	"path"
+	"strings"
 	"time"
 
 	validation "github.com/go-ozzo/ozzo-validation/v4"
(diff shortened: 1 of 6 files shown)


kjaskiewiczz:

model/image.go
```diff
@@ -31,8 +32,19 @@ import (
...
+	runeDollar         = '\uFF04'
+	runeDot            = '\uFF0E'
 )
 
+var (
+	StorageKeyImageProvidesRootFSChecksum = "meta_artifact.provides." +
+		GetProvidesKeyReplacer().Replace("rootfs-image.checksum")
+	StorageKeyImageProvidesRootFSVersion = "meta_artifact.provides." +
+		GetProvidesKeyReplacer().Replace("rootfs-image.version")
+)
```
tranchitella, line 44: Not tested, but you get the idea... why initializing the replacer each time you use it?

```suggestion
var (
        providesKeyReplacer = GetProvidesKeyReplacer()
	StorageKeyImageProvidesRootFSChecksum = "meta_artifact.provides." +
		providesKeyReplacer.Replace("rootfs-image.checksum")
	StorageKeyImageProvidesRootFSVersion = "meta_artifact.provides." +
		providesKeyReplacer.Replace("rootfs-image.version")
)
```

model/image.go
```diff
@@ -267,3 +279,18 @@ func (msg MultipartGenerateImageMsg) Validate() error {
...
+
+func GetProvidesKeyReplacer() *strings.Replacer {
+	return strings.NewReplacer(".", string(runeDot), "$", string(runeDollar))
+}
+
+// MarshalBSONValue marshals the Provides to a mongo-compatible
+// document.
+func (p Provides) MarshalBSONValue() (bsontype.Type, []byte, error) {
+	attrs := make(bson.M, len(p))
+	replacer := GetProvidesKeyReplacer()
```
tranchitella, line 291:
```suggestion
```

model/image.go
```diff
@@ -267,3 +279,18 @@ func (msg MultipartGenerateImageMsg) Validate() error {
...
+	for k, v := range p {
+		attrs[replacer.Replace(k)] = v
```
tranchitella, line 293:
```suggestion
		attrs[providesKeyReplacer.Replace(k)] = v
```

model/image.go
kjaskiewiczz: done

model/image.go
kjaskiewiczz: done

model/image.go
kjaskiewiczz: done (the last commit)

store/mongo/datastore_mongo.go
```diff
@@ -325,6 +328,20 @@ var (
...
+	IndexArtifactProvidesRootfsVersionAndChecksum = mongo.IndexModel{
+		Keys: bson.D{
+			{Key: model.StorageKeyImageProvidesRootFSVersion,
+				Value: 1},
+			{Key: model.StorageKeyImageProvidesRootFSChecksum,
+				Value: 1},
+		},
+		Options: &mopts.IndexOptions{
+			Background: &_false,
+			Name:       &IndexArtifactProvidesRootfsVersionAndChecksumName,
```
alfrunes, line 342: This index needs to be sparse or use a `PartialFilterExpression` to cover documents without Provides.

kjaskiewiczz:
> The index is not compatible with the new schema.

not sure I get it; I would appreciate an explanation for dummies ;)

> General question: Since we're migrating all documents, maybe we should consider changing the schema for provides to hold embedded documents of key/value pairs? It might make the index a bit fatter, but it adds the ability to query arbitrary provides values if there should be a need.

are embedded documents in depends/provides supported by the mender-artifact?

kjaskiewiczz:
> I think we impose that provides/depends needs to be a flat string mapping. In that case, my question is if we should encode these key/value pairs as embedded documents in MongoDB on the form:
> 
> ```go
> {"provides": [{"key": "<provides1>", "value": "value1"},{"key": "<provides2>", "value": "<value2>"}, ...]}
> ```

currently I'm creating index with rootfs-image.checksum and rootfs-image.versin only; with embedded documents I'd have to index all provides, right (index for provides.key)? 
the index will potentially be quite big, but it would solve [this issue](https://tracker.mender.io/browse/MEN-6063?focusedCommentId=128669&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-128669)

store/mongo/datastore_mongo.go
kjaskiewiczz: it's a regular index (not unique), so, as you pointed out, there is no need for PartialFilterExpression

tranchitella (changes requested):
Let's close this PR in favour of https://github.com/mendersoftware/deployments/pull/796

PR #796, opened 2022-12-08
feat: change image (aka artifact) schema

kjaskiewiczz:
With the new schema it will be possible to query by artifact provides.
Ticket: [MEN-6107](https://tracker.mender.io/browse/MEN-6107)

### DIFF
--- a/model/image.go
+++ b/model/image.go
@@ -33,6 +33,13 @@ const (
 	ArtifactFileSuffix = ".mender"
 )
 
+var (
+	StorageKeyImageProvidesKey   = "meta_artifact.provides.key"
+	StorageKeyImageProvidesValue = "meta_artifact.provides.value"
+)
+
+type Provides map[string]string
+
 func ImagePathFromContext(ctx context.Context, id string) string {
 	imgPath := id
 	if idty := identity.FromContext(ctx); idty != nil {
@@ -95,9 +102,9 @@ type ArtifactMeta struct {
 	// List of updates
 	Updates []Update `json:"updates" valid:"-"`
 
-	// Provides is a map[string]interface{} (JSON) of artifact_provides used
+	// Provides is a map of artifact_provides used
 	// for checking artifact (version 3) dependencies.
-	Provides map[string]string `json:"artifact_provides,omitempty" bson:"provides" valid:"-"`
+	Provides Provides `json:"artifact_provides,omitempty" bson:"provides,omitempty" valid:"-"`
 
 	// Depends is a map[string]interface{} (JSON) of artifact_depends used
 	// for checking/validate against artifact (version 3) provides.
@@ -265,5 +272,40 @@ func (msg MultipartGenerateImageMsg) Validate() error {
 	if msg.FileReader == nil {
 		return errors.New("missing 'file' section")
 	}
+	return nil
+}
+
+type provideInternal struct {
+	Key   string
+	Value string
+}
+
+func (p Provides) MarshalBSONValue() (bsontype.Type, []byte, error) {
+	attrs := make([]provideInternal, len(p))
+	i := 0
+	for k, v := range p {
+		attrs[i].Key = k
+		attrs[i].Value = v
+		i++
+	}
+	return bson.MarshalValue(attrs)
+}
+
+func (p *Provides) UnmarshalBSONValue(t bsontype.Type, b []byte) error {
+	raw := bson.Raw(b)
+	elems, err := raw.Elements()
+	if err != nil {
+		return err
+	}
+	*p = make(Provides, len(elems))
+	var tmp provideInternal
+	for _, elem := range elems {
+		err = elem.Value().Unmarshal(&tmp)
+		if err != nil {
+			return err
+		}
+		(*p)[tmp.Key] = tmp.Value
+	}
+
 	return nil
 }
--- a/store/mongo/datastore_mongo.go
+++ b/store/mongo/datastore_mongo.go
@@ -80,6 +80,9 @@ var (
 	// Indexes 1.2.6
 	IndexDeviceDeploymentStatusName = "deploymentid_status_deviceid"
 
+	// Indexes 1.2.13
+	IndexArtifactProvidesName = "artifact_provides"
+
 	_false         = false
 	_true          = true
 	StorageIndexes = mongo.IndexModel{
@@ -325,6 +328,21 @@ var (
 		Options: mopts.Index().
 			SetName(IndexDeviceDeploymentsLogs),
 	}
+
+	// 1.2.13
+	IndexArtifactProvides = mongo.IndexModel{
+		Keys: bson.D{
+			{Key: model.StorageKeyImageProvidesKey,
+				Value: 1},
+			{Key: model.StorageKeyImageProvidesValue,
+				Value: 1},
+		},
+		Options: &mopts.IndexOptions{
+			Background: &_false,
+			Sparse:     &_true,
+			Name:       &IndexArtifactProvidesName,
+		},
+	}
 )
 
 // Errors
@@ -358,6 +376,7 @@ const (
 	// Need to be kept in sync with structure filed names
 	StorageKeyId = "_id"
 
+	StorageKeyImageProvides    = "meta_artifact.provides"
 	StorageKeyImageDepends     = "meta_artifact.depends"
 	StorageKeyImageDependsIdx  = "meta_artifact.depends_idx"
 	StorageKeyImageSize        = "size"
--- a/store/mongo/migration_1_2_13.go
+++ b/store/mongo/migration_1_2_13.go
@@ -1,0 +1,139 @@
+// Copyright 2022 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+package mongo
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/mendersoftware/go-lib-micro/mongo/migrate"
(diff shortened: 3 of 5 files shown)


kjaskiewiczz:

store/mongo/datastore_mongo.go
```diff
@@ -325,6 +328,21 @@ var (
...
 			SetName(IndexDeviceDeploymentsLogs),
 	}
+
+	// 1.2.13
+	IndexArtifactProvides = mongo.IndexModel{
+		Keys: bson.D{
+			{Key: model.StorageKeyImageProvidesKey,
+				Value: 1},
+			{Key: model.StorageKeyImageProvidesValue,
+				Value: 1},
```
tranchitella, line 338: Which queries will match this index?
Please note that you are indexing key and value separately, and not combined.
In order to query for specific key/vaule we'll have to use $elemMatch.

model/image.go
```diff
@@ -267,3 +274,38 @@ func (msg MultipartGenerateImageMsg) Validate() error {
...
 }
+
+type provideInternal struct {
+	Key   string
+	Value string
+}
+
+func (p Provides) MarshalBSONValue() (bsontype.Type, []byte, error) {
+	attrs := make([]provideInternal, len(p))
+	i := 0
```
tranchitella, line 285: What is i used for?

store/mongo/datastore_mongo.go
kjaskiewiczz: the query will be in the enterprise version of the repo;
the idea [there](https://github.com/mendersoftware/deployments/pull/795#issuecomment-1340510348) was to store provides as a list of embedded documents, and yes, in this case the query will use $elemMatch (I should probably use it in the test, so it would more clear how I'm planning to use it)

model/image.go
kjaskiewiczz: `i`? to iterate over attrs?

store/mongo/migration_1_2_13.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// Depends is a map[string]interface{} (JSON) of artifact_depends used
+	// for checking/validate against artifact (version 3) provides.
+	Depends map[string]interface{} `json:"artifact_depends,omitempty" bson:"depends" valid:"-"`
+
+	// ClearsProvides is a list of strings (JSON) of clears_artifact_provides used
+	// for clearing already-installed artifact (version 3) provides.
+	//nolint:lll
+	ClearsProvides []string `json:"clears_artifact_provides,omitempty" bson:"clears_provides,omitempty" valid:"-"`
+}
```
alfrunes, line 83: nit: Only use the `Id` and `Provides` fields are used in this struct, the rest can be stripped away.

alfrunes (commented):
Just an additional remark about this approach: You cannot use the index for more than one key/value ($elemMatch) query criteria. You can of course combine multiple $elemMatch on the same field using $and, but only one of the queries are backed by the index (IXSCAN). By chaining multiple queries this way, it does not seem to be deterministic which filter condition is actually backed by the index.

kjaskiewiczz:
> Just an additional remark about this approach: You cannot use the index for more than one key/value ($elemMatch) query criteria. You can of course combine multiple $elemMatch on the same field using $and, but only one of the queries are backed by the index (IXSCAN). By chaining multiple queries this way, it does not seem to be deterministic which filter condition is actually backed by the index.

```
db.images.find({$and: [{"meta_artifact.provides": {$elemMatch: {"key":"p3", "value":"v3"}}}, {"meta_artifact.provides": {$elemMatch: {"key":"p1", "value":"v1"}}}]}).pretty()
{
	"_id" : "1c7e7f59-50f0-4f7c-afe1-ab8fa3ff2bf4",
	"meta" : {
		"description" : ""
	},
	"meta_artifact" : {
		"name" : "test",
		"device_types_compatible" : [
			"foo"
		],
		"info" : {
			"format" : "mender",
			"version" : NumberLong(3)
		},
		"signed" : false,
		"updates" : [
			{
				"typeinfo" : {
					"type" : null
				},
				"files" : null,
				"metadata" : null
			}
		],
		"provides" : [
			{
				"key" : "p2",
				"value" : "v2"
			},
			{
				"key" : "p3",
				"value" : "v3"
			},
			{
				"key" : "p4",
				"value" : "v4"
			},
			{
				"key" : "artifact_name",
				"value" : "test"
			},
			{
				"key" : "foo",
				"value" : "bar"
			},
			{
				"key" : "p1",
				"value" : "v1"
			}
		],
		"depends" : {
			"device_type" : [
				"foo"
			]
		},
		"depends_idx" : [
			{
				"device_type" : "foo"
			}
		]
	},
	"size" : NumberLong(5120),
	"modified" : ISODate("2022-12-09T14:30:21.969Z")
}

db.images.find({$and: [{"meta_artifact.provides": {$elemMatch: {"key":"p3", "value":"v3"}}}, {"meta_artifact.provides": {$elemMatch: {"key":"p1", "value":"v1"}}}]}).explain()
{
	"queryPlanner" : {
		"plannerVersion" : 1,
		"namespace" : "deployment_service.images",
		"indexFilterSet" : false,
		"parsedQuery" : {
			"$and" : [
				{
					"meta_artifact.provides" : {
						"$elemMatch" : {
							"$and" : [
								{
									"key" : {
										"$eq" : "p3"
									}
								},
								{
									"value" : {
										"$eq" : "v3"
									}
								}
							]
						}
					}
				},
				{
					"meta_artifact.provides" : {
						"$elemMatch" : {
							"$and" : [
								{
									"key" : {
										"$eq" : "p1"
									}
								},
								{
									"value" : {
										"$eq" : "v1"
									}
								}
							]
						}
					}
				}
			]
		},
		"queryHash" : "731816BF",
		"planCacheKey" : "731816BF",
		"winningPlan" : {
			"stage" : "FETCH",
			"filter" : {
				"$and" : [
					{
						"meta_artifact.provides" : {
							"$elemMatch" : {
								"$and" : [
									{
										"key" : {
											"$eq" : "p3"
										}
									},
									{
										"value" : {
											"$eq" : "v3"
										}
									}
								]
							}
						}
					},
					{
						"meta_artifact.provides" : {
							"$elemMatch" : {
								"$and" : [
									{
										"key" : {
											"$eq" : "p1"
										}
									},
									{
										"value" : {
											"$eq" : "v1"
										}
									}
								]
							}
						}
					}
				]
			},
			"inputStage" : {
				"stage" : "IXSCAN",
				"keyPattern" : {
					"meta_artifact.provides.key" : 1,
					"meta_artifact.provides.value" : 1
				},
				"indexName" : "artifact_provides",
				"isMultiKey" : true,
				"multiKeyPaths" : {
					"meta_artifact.provides.key" : [
						"meta_artifact.provides"
					],
					"meta_artifact.provides.value" : [
						"meta_artifact.provides"
					]
				},
				"isUnique" : false,
				"isSparse" : true,
				"isPartial" : false,
				"indexVersion" : 2,
				"direction" : "forward",
				"indexBounds" : {
					"meta_artifact.provides.key" : [
						"[\"p3\", \"p3\"]"
					],
					"meta_artifact.provides.value" : [
						"[\"v3\", \"v3\"]"
					]
				}
			}
		},
		"rejectedPlans" : [
			{
				"stage" : "FETCH",
				"filter" : {
					"$and" : [
						{
							"meta_artifact.provides" : {
								"$elemMatch" : {
									"$and" : [
										{
											"key" : {
												"$eq" : "p1"
											}
										},
										{
											"value" : {
												"$eq" : "v1"
											}
										}
									]
								}
							}
						},
						{
							"meta_artifact.provides" : {
								"$elemMatch" : {
									"$and" : [
										{
											"key" : {
												"$eq" : "p3"
											}
										},
										{
											"value" : {
												"$eq" : "v3"
											}
										}
									]
								}
							}
						}
					]
				},
				"inputStage" : {
					"stage" : "IXSCAN",
					"keyPattern" : {
						"meta_artifact.provides.key" : 1,
						"meta_artifact.provides.value" : 1
					},
					"indexName" : "artifact_provides",
					"isMultiKey" : true,
					"multiKeyPaths" : {
						"meta_artifact.provides.key" : [
							"meta_artifact.provides"
						],
						"meta_artifact.provides.value" : [
							"meta_artifact.provides"
						]
					},
					"isUnique" : false,
					"isSparse" : true,
					"isPartial" : false,
					"indexVersion" : 2,
					"direction" : "forward",
					"indexBounds" : {
						"meta_artifact.provides.key" : [
							"[\"p1\", \"p1\"]"
						],
						"meta_artifact.provides.value" : [
							"[\"v1\", \"v1\"]"
						]
					}
				}
			}
		]
	},
	"serverInfo" : {
		"host" : "b5d96ac267c1",
		"port" : 27017,
		"version" : "4.4.17",
		"gitVersion" : "85de0cc83f4dc64dbbac7fe028a4866228c1b5d1"
	},
	"ok" : 1
}
```

maybe I misunderstood your comment, but the input stage here is `IXSCAN`, so we should be fine, right?

alfrunes:
> maybe I misunderstood your comment, but the input stage here is `IXSCAN`, so we should be fine, right?

My comment is that it's running a filter on top of an IXSCAN, which of the two $and terms that is used for the IXSCAN seems to be random, and there seems to be no way of controlling which one is used.

tranchitella (approved):

tranchitella:
@alfrunes I am not sure I understand, to me it looks like the index is correctly used:

```
				"indexBounds" : {
					"meta_artifact.provides.key" : [
						"[\"p3\", \"p3\"]"
					],
					"meta_artifact.provides.value" : [
						"[\"v3\", \"v3\"]"
					]
				}
```

kjaskiewiczz:
@alfrunes in the deployments-enterprise, the query will also use device type, but since we store this information also in an array, I cannot create index with both, depends and device type (it's not possible to combine multikey indexes);
so all I could come up with is what you can see in this PR

alfrunes:
> @alfrunes I am not sure I understand, to me it looks like the index is correctly used:
> 
> ```
> 				"indexBounds" : {
> 					"meta_artifact.provides.key" : [
> 						"[\"p3\", \"p3\"]"
> 					],
> 					"meta_artifact.provides.value" : [
> 						"[\"v3\", \"v3\"]"
> 					]
> 				}
> ```

The thing is it seems to be random which of the winningPlan or the rejecedPlan is selected, i.e. which term is selected for defining index bounds and which filters the matching documents.

alfrunes (approved):
I won't block this PR, but let's be mindful about using provides. I believe we can fall back to aggregation pipelines for forcing the desired index usage if required.

PR #798, opened 2022-12-11
feat: use scratch as base image, run as non-root user

tranchitella:
Changelog: Title
Ticket: MEN-5658


tranchitella:

storage/certs.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"crypto/x509"
+	"os"
+)
+
+const storageBackendCertEnvName = "STORAGE_BACKEND_CERT"
+
+var getEnv = func(key string) string {
+	return os.Getenv(key)
+}
```
alfrunes, line 26: I don't get why this is needed, the closure has same call signature as `os.Getenv`.

alfrunes (approved):

storage/certs.go
tranchitella: I replace it in unit tests.

PR #802, opened 2022-12-15
follow up from autogenerate delta testing

kjaskiewiczz:


kjaskiewiczz:

app/app.go
```diff
@@ -1781,3 +1788,29 @@ func (d *Deployments) search(
...
+	artifactName string,
+) error {
+	// first check if there are pending deployments with given artifact name
+	exists, err := d.db.ExistUnfinishedByArtifactName(ctx, artifactName)
+	if err != nil {
+		return errors.Wrap(err, "looking for deployments with given artifact name")
+	}
+	if !exists {
+		return nil
+	}
```
tranchitella, line 1803: I would write it as `else if`; I'm not asking you to change it, just sharing. :)

store/mongo/datastore_mongo.go
```diff
@@ -2354,6 +2354,40 @@ func (db *DataStoreMongo) ExistUnfinishedByArtifactId(ctx context.Context,
 	return true, nil
 }
 
+// ExistUnfinishedByArtifactName checks if there is an active deployment that uses
+// given artifact
```
tranchitella, line 2358: *the* given artifact

tranchitella (approved):

app/app.go
```diff
@@ -1781,3 +1788,29 @@ func (d *Deployments) search(
...
+
+	// Assign artifacts to the deployments with given artifact name
+	artifacts, err := d.db.ImagesByName(ctx, artifactName)
+	if err != nil {
+		return errors.Wrap(err, "Finding artifact with given name")
+	}
+
+	if len(artifacts) == 0 {
+		return ErrNoArtifact
+	}
```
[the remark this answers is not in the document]
kjaskiewiczz: this is actually interesting; I've been asked to use `else` couple of times, but I can't get a habit of using it;
I think it's mainly because I've been learned to not use it back in a days when I was learning C in my first job;
I'm not an expert, and I'm not a fan of such discussions, but there are some patterns and discussions around usage of `else` statement e.g. it's part of the [return early pattern](https://medium.com/swlh/return-early-pattern-3d18a41bba8);
and I'm also just sharing :) (really)

PR #803, opened 2022-12-18
MEN-6134, MEN-6135: add support for device deployments history deletion

tranchitella:


docs/management_api.yml
```diff
@@ -693,6 +693,38 @@ paths:
...
+      tags:
+        - Management API
+      security:
+        - ManagementJWT: []
+      summary: Delete the Device Deployments history
+      description: |
+        Delete the completed Device Deployments records for the given device.
+        This effectively resets the Device Deployments history and makes the
+        device eligible (again) for all the active deployments.
+      consumes:
```
kjaskiewiczz, line 708: I don't think it consumes mulitpart/form-data

kjaskiewiczz (changes requested):
only small change in the docs is needed

tranchitella:

store/mongo/datastore_mongo.go
```diff
@@ -1715,6 +1720,32 @@ func (db *DataStoreMongo) AbortDeviceDeployments(ctx context.Context,
...
+	}
+
+	now := time.Now()
+	update := bson.M{
+		"$set": bson.M{
+			StorageKeyDeviceDeploymentDeleted: &now,
+		},
+	}
+
+	if _, err := collDevs.UpdateMany(ctx, selector, update); err != nil {
```
kjaskiewiczz, line 1742: doesn't really matter, but why `UpdateMany` insead of `UpdateOne`?

app/app.go
```diff
@@ -1194,22 +1201,23 @@ func (d *Deployments) createDeviceDeploymentWithStatus(
 		return nil, err
 	}
 
-	if err := d.db.InsertDeviceDeployment(ctx, deviceDeployment); err != nil {
+	if err := d.db.InsertDeviceDeployment(ctx, deviceDeployment,
```
kjaskiewiczz, line 1204: just for my knowledge - the approach here is to not decrement deployment counter when "deleting" device deployment, but to not increment it, if the device will become a part of the same deployment once again? not sure if this will work in all cases (e.g. after resetting the device, device will not ask for the deployment), but it's probably OK (I think someone who will use this functionality will know what he's doing);
it may be more complex with phased rollouts in the deployments-enterprise, but I'm just guessing here;

kjaskiewiczz (changes requested):

store/mongo/datastore_mongo.go
tranchitella: Because this will (logically) delete multiple records, in particular all the device deployment records for a given device that are NOT active (historical data).

app/app.go
tranchitella: Correct, we don't decrement the device count in the deployments when we (logically) remove the device deployment records. This is because that counter determines whether the deployment is completed, and this has several side effects. All in all, the data is there, we just mark it as "deleted" so we can ignore it when the device asks for updates, again. This is relevant for dynamic deployments, while it has no practical implications for static deployments. I tested it (manually) bringing these changes to deployments-enterprise and it works correctly.

I want to add that my idea is that we should refactor the retry feature in deployments-enterprise to take advantage of the `deleted` device deployment field, creating a new entry for each try. This will show the different attempts as history independent items in the device deployments history view we are implementing.

store/mongo/datastore_mongo.go
kjaskiewiczz: right; sorry for the noise

kjaskiewiczz (approved):
looks good

alfrunes (approved):

PR #809, opened 2022-12-23
MEN-5911: feat: internal end-point to retrieve the device deployments history

tranchitella:
feat: internal end-point to retrieve the device deployments history

Changelog: Title
Ticket: MEN-5911


tranchitella:
@mender-test-bot start pipeline --pr reporting/86

tranchitella:
@mender-test-bot start pipeline --pr reporting/86 --pr deployments-enterprise/745

store/mongo/datastore_mongo.go
```diff
@@ -1530,6 +1530,20 @@ func (db *DataStoreMongo) GetDevicesListForDeployment(ctx context.Context,
 					Value: model.DeviceDeploymentStatusPending,
 				}},
 			})
+		} else if *q.Status == model.DeviceDeploymentStatusFinishedStr {
```
kjaskiewiczz, line 1533: we have `active` and `finished` fields in the device deployment object, maybe we could use them to create a query for active and finished device deployments?

kjaskiewiczz (approved):
only small comment

store/mongo/datastore_mongo.go
tranchitella: I believe the reason why we query by status is because it uses the index in MongoDB.
Querying by other fields requires additional indexes.

tranchitella:
@mender-test-bot start pipeline --pr reporting/86 --fast

store/mongo/datastore_mongo.go
kjaskiewiczz: you're right; we don't have indexes for `active` and `finished`;
I thought we introduced `active` to improve some queries and we introduced index with that field, but, after checking, looks like it's not the case;
sorry for the noise;

tranchitella:
@mender-test-bot start pipeline --pr deployments-enterprise/745 --pr reporting/86 --pr integration/2299

tranchitella:
Pipeline is here: https://github.com/mendersoftware/integration/pull/2299

PR #810, opened 2022-12-30
MEN-5930: index device deployment objects

tranchitella:


tranchitella:
@mender-test-bot start pipeline --pr reporting/87

app/app.go
```diff
@@ -1264,47 +1264,16 @@ func (d *Deployments) isDevicePartOfDeployment(
...
-			}
-			return nil, ErrConflictingRequestData
-		}
-	} else {
-		// save the request
-		if err := d.db.SaveDeviceDeploymentRequest(
-			ctx, deviceDeployment.Id, request); err != nil {
-			return nil, err
-		}
+	err = d.saveDeviceDeploymentRquest(ctx, deviceID, deviceDeployment, request)
```
alfrunes, line 1274:
```suggestion
	err = d.saveDeviceDeploymentRequest(ctx, deviceID, deviceDeployment, request)
```

app/app.go
```diff
@@ -1394,6 +1364,45 @@ func (d *Deployments) GetDeploymentForDeviceWithCurrent(ctx context.Context, dev
...
+			// mark deployment for this device as failed to force client to rollback
+			l := log.FromContext(ctx)
+			l.Errorf(
+				"Device with id %s reported new data: %s during update process;"+
+					"old data: %s",
+				deviceID, request, deviceDeployment.Request)
+
+			if err := d.UpdateDeviceDeploymentStatus(ctx, deviceDeployment.DeploymentId, deviceID,
+				model.DeviceDeploymentState{
+					Status: model.DeviceDeploymentStatusFailure,
```
alfrunes, line 1382: Random observation: I feel like we should set the SubState here to make it obvious from the UI what happened.

app/app.go
```diff
@@ -1741,11 +1765,16 @@ func (d *Deployments) updateDeviceDeploymentsStatus(
...
+				deviceDeployment, err := d.createDeviceDeploymentWithStatus(ctx,
 					deviceId, deployment, status)
 				if err != nil {
 					return err
 				}
+				if err := d.reindexDeployment(ctx, deviceDeployment.DeviceId,
+					deviceDeployment.DeploymentId, deviceDeployment.Id); err != nil {
+					l := log.FromContext(ctx)
+					l.Warn(errors.Wrap(err, "failed to trigger a device reindex"))
+				}
```
alfrunes, line 1786: Shouldn't there be a condition on the status being a terminal state (`success`, `failure`, `already-installed`, `noartifact`, `decommissioned` or `aborted`)?

app/app.go
tranchitella: The idea is valuable, but I don't think it fits into my PR.

app/app.go
tranchitella: you are right

PR #813, opened 2023-01-12
feat: add special representation of artifact provides to artifact object

kjaskiewiczz:
This way we can index provides, but we don't have to wait for the migration to complete before we'll start using new index.


kjaskiewiczz:

tranchitella (approved):
Good idea!

alfrunes (approved):

PR #816, opened 2023-01-24
feat: Upgrade to mender-artifact with zstd compression support.

kacf:
Changelog: Add support for zstd compressed artifacts. Ticket: None

Signed-off-by: Kristian Amlie <kristian.amlie@northern.tech>


kacf:
Nothing else is needed to enable zstd compression support in the service, is it? It should be detected by extension.

alfrunes (approved):

PR #835, opened 2023-03-06
GetDeploymentForDeviceWithCurrent refactoring + fix deployment total count calculation

kjaskiewiczz:
Ticket: [MEN-6363](https://northerntech.atlassian.net/browse/MEN-6363)

[MEN-6363]: https://northerntech.atlassian.net/browse/MEN-6363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1335,6 +1335,7 @@ func (d *Deployments) GetDeploymentForDeviceWithCurrent(ctx context.Context, dev
 			l := log.FromContext(ctx)
 			l.Warn(errors.Wrap(err, "failed to trigger a device reindex"))
 		}
+		var newArtifactAssigned bool
 
 		return nil, nil
 	}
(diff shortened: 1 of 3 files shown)


app/app.go
```diff
@@ -1335,6 +1335,7 @@ func (d *Deployments) GetDeploymentForDeviceWithCurrent(ctx context.Context, dev
 			l := log.FromContext(ctx)
 			l.Warn(errors.Wrap(err, "failed to trigger a device reindex"))
 		}
+		var newArtifactAssigned bool
```
tranchitella: Is this `var` in the right place? Why is it just before the `return nil, nil`?

kjaskiewiczz: it's at the beginning of the method; maybe it's not that well visible in the unified view?

tranchitella: Yep, it was a glitch in the github view.

tranchitella (approved):

PR #837, opened 2023-03-08
feat: omit configuration values in the deployment details

0lmi:
Configuration settings can contain sensitive values. Therefore, we mask the values to limit the reask of leaking secrets through the deployments APIs.

Changelog: Title
Ticket: None


0lmi:

kacf (commented):
Normally `feat:` fixes are not put in patch releases, but I suppose this is a special case because there is risk of leaking data.

I approve, but I leave the "official" approval to one of the backend guys, since I don't know this code myself.

tranchitella (approved):

PR #838, opened 2023-03-09
fix(list device deployments): device ID is not always UUID

kjaskiewiczz:
Ticket: [MEN-6376](https://northerntech.atlassian.net/browse/MEN-6376)

[MEN-6376]: https://northerntech.atlassian.net/browse/MEN-6376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/http/api_deployments.go
+++ b/api/http/api_deployments.go
@@ -95,6 +95,7 @@ const (
 // Errors
 var (
 	ErrIDNotUUID                      = errors.New("ID is not a valid UUID")
+	ErrEmptyID                        = errors.New("id: cannot be blank")
 	ErrArtifactUsedInActiveDeployment = errors.New("Artifact is used in active deployment")
 	ErrInvalidExpireParam             = errors.New("Invalid expire parameter")
 	ErrArtifactNameMissing            = errors.New(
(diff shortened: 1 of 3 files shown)


api/http/api_deployments.go
```diff
@@ -1741,16 +1742,12 @@ func (d *DeploymentsApiHandlers) listDeviceDeployments(ctx context.Context,
 	var IDs []string
 	if byDeviceID {
 		did = r.PathParam("id")
-		if !govalidator.IsUUID(did) {
-			d.view.RenderError(w, r, ErrIDNotUUID, http.StatusBadRequest, l)
-			return
-		}
```
merlin-northern, line 1747: we know that it is either UUID (in `govalidator.IsUUID` understanding) or the legacy id. we should add the validation for the latter format here rather than removing current one, and extend the tests accordingly.

kjaskiewiczz: sorry for the stupid question, but what this check will actually give us?

tranchitella (approved):

PR #841, opened 2023-03-14
MEN-6337: Direct artifact upload APIs

alfrunes:


alfrunes:

docs/management_api.yml
```diff
@@ -1197,9 +1197,36 @@ paths:
         200:
           description: OK
           schema:
-            type: object
-            properties:
-              $ref: "#/definitions/ArtifactUploadLink"
+            $ref: "#/definitions/ArtifactUploadLink"
+        400:
```
kjaskiewiczz, line 1201: how can I get 400 when using this endpoint? isn't it the POST without body?
and, btw, why it's POST and not GET?

kjaskiewiczz (commented):
If I understand correctly, the complete operation is asynchronous (I think we originally planned to have it synchronous), which is OK, but is it possible to check the status of processing (and see error if it will fail?)?

api/http/api_deployments.go
merlin-northern, line 1: note to self https://github.com/mendersoftware/mender-cli/pull/182 has to call the complete endpoint.

docs/management_api.yml
alfrunes: I updated the swagger documentation reflecting the actual responses from the API.
It's `POST` and not `GET` because you create a "Link" object.

alfrunes:
> If I understand correctly, the complete operation is asynchronous (I think we originally planned to have it synchronous), which is OK, but is it possible to check the status of processing (and see error if it will fail?)?

At this point there is not endpoint for retrieving the status for an upload in progress as this was not part of the requirements. However, I made the document structure in such a way that it should be fairly straight forward to make such and API, but I would put that in a separate ticket since this and the [following ticket](https://northerntech.atlassian.net/browse/MEN-6339) is already beyond the estimated effort of the tickets.

kjaskiewiczz:
As I mentioned before, I think the original plan for /complete endpoint was to have it synchronous. Since now the operation is asynchronous, I think we need at least a follow up ticket with an endpoint for checking the status. Can you create one?

kjaskiewiczz (approved):
good job!

PR #846, opened 2023-03-25
Dependabot updates

tranchitella:


merlin-northern (approved):
as usual I did: `rm -Rf vendor; go mod tidy; go mod vendor; git log --stat=1024 -4 | grep -v '^[ ]*vendor'; git status;` and then looked at the commits as well.

PR #848, opened 2023-03-27
fix(docs): add undocumented groups field to deployment object

kjaskiewiczz:


alfrunes (approved):

docs/management_api.yml
```diff
@@ -1597,6 +1597,13 @@ definitions:
         description: An array of artifact's identifiers.
         items:
           type: string
+      groups:
+        type: array
+        description: |
+            An array of groups the devices targeted by the deployment belongs to.
```
tranchitella, line 1603:
```suggestion
            An array of groups the devices targeted by the deployment belong to.
```

docs/management_api.yml
```diff
@@ -1597,6 +1597,13 @@ definitions:
...
+            Available only if user created deployment for a group or for a single device (if the device was in a static group).
```
tranchitella, line 1604:
```suggestion
            Available only if the user created the deployment for a group or a single device (if the device was in a static group).
```

tranchitella (approved):
Pre-approving, the change is trivial.

docs/management_api.yml
kjaskiewiczz: fixed

docs/management_api.yml
kjaskiewiczz: fixed

PR #849, opened 2023-04-11
feat: Store device deployments status separately

merlin-northern:
feat: Store device deployments status separately

Changelog: Title
Ticket: [MEN-6423](https://northerntech.atlassian.net/browse/MEN-6423)
Signed-off-by: Peter Grzybowski <peter@northern.tech>
[MEN-6423](https://northerntech.atlassian.net/browse/MEN-6423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ)
[goes with pull/850](https://github.com/mendersoftware/deployments/pull/850)

[MEN-6423]: https://northerntech.atlassian.net/browse/MEN-6423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MEN-6423]: https://northerntech.atlassian.net/browse/MEN-6423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #850, opened 2023-04-14
feat: internal endpoint to get the latest device deployment statuses

merlin-northern:
feat: internal endpoint to get the latest device deployment statuses

Changelog: Title
Ticket: MEN-6421
[goes with pull/849](https://github.com/mendersoftware/deployments/pull/849)

### DIFF
--- a/api/http/api_deployments_last_status.go
+++ b/api/http/api_deployments_last_status.go
@@ -1,0 +1,71 @@
+// Copyright 2023 Northern.tech AS
+//
+//	Licensed under the Apache License, Version 2.0 (the "License");
+//	you may not use this file except in compliance with the License.
+//	You may obtain a copy of the License at
+//
+//	    http://www.apache.org/licenses/LICENSE-2.0
+//
+//	Unless required by applicable law or agreed to in writing, software
+//	distributed under the License is distributed on an "AS IS" BASIS,
+//	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//	See the License for the specific language governing permissions and
+//	limitations under the License.
+
+package http
+
+import (
+	"net/http"
+
+	"github.com/pkg/errors"
+
+	"github.com/ant0ine/go-json-rest/rest"
+
+	"github.com/mendersoftware/go-lib-micro/identity"
+	"github.com/mendersoftware/go-lib-micro/requestlog"
+	"github.com/mendersoftware/go-lib-micro/rest_utils"
+)
+
+// device deployments last status handler
+func (d *DeploymentsApiHandlers) GetDeviceDeploymentLastStatus(
+	w rest.ResponseWriter,
+	r *rest.Request,
+) {
+	l := requestlog.GetRequestLogger(r)
+
+	l.Debugf("starting")
+
+	tenantId := r.PathParam("tenant")
+	var devicesIds []string
+	if err := r.DecodeJsonPayload(&devicesIds); err != nil {
+		l.Errorf("error during DecodeJsonPayload: %s.", err.Error())
+		rest_utils.RestErrWithLog(
+			w,
+			r,
+			l,
+			errors.Wrap(err, "cannot parse device ids array"),
+			http.StatusBadRequest,
+		)
+		return
+	}
+
+	l.Debugf("querying %d devices ids", len(devicesIds))
+	ctx := r.Context()
+	if tenantId != "" {
+		ctx = identity.WithContext(
+			ctx,
+			&identity.Identity{
+				Tenant: tenantId,
+			},
+		)
+	}
+	lastDeployments, err := d.app.GetDeviceDeploymentLastStatus(ctx, devicesIds)
+	switch err {
+	default:
+		d.view.RenderInternalError(w, r, err, l)
+	case nil:
+		l.Infof("outputting: %+v", lastDeployments)
+		w.WriteHeader(http.StatusOK)
+		_ = w.WriteJson(lastDeployments)
+	}
+}
--- a/docs/internal_api.yml
+++ b/docs/internal_api.yml
@@ -512,6 +512,38 @@ paths:
         500:
           $ref: "#/responses/InternalServerError"
 
+  /tenants/{tenant_id}/devices/deployments/last:
+    post:
+      operationId: Get last device deployment status
+      tags:
+        - Internal API
+      summary: Get status of the last device devployment
+      description: |
+        Return the status of the last unsucessful device deployment.
+      parameters:
+        - name: tenant_id
+          in: path
+          description: Tenant identifier.
+          required: true
+          type: string
+        - name: devicesIds
+          in: body
+          description: An array of device ids for which to return the statuses.
+          schema:
+            $ref: "#/definitions/LastDeviceDeploymentsStatuses"
+          required: true
+      consumes:
+        - application/json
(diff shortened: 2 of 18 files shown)


merlin-northern:

merlin-northern:
@kjaskiewiczz units in app and api are missing but please give me an early review so I can add them and not rewrite 4 times ;-) the acceptance tests I added.

docs/internal_api.yml
```diff
@@ -512,6 +512,38 @@ paths:
         500:
           $ref: "#/responses/InternalServerError"
 
+  /device/deployments/last:
```
kjaskiewiczz, line 515: the path is like nothing else in this service, but maybe it's OK; I don't better idea anyway...

kjaskiewiczz: where we'll take the tenant id from?

store/mongo/datastore_mongo.go
```diff
@@ -2774,3 +2774,25 @@ func (db *DataStoreMongo) SaveLastDeviceDeploymentStatus(
 	}
 	return err
 }
+
+func (db *DataStoreMongo) GetLastDeviceDeploymentStatus(
+	ctx context.Context,
+	devicesIds []string,
+) ([]model.DeviceDeploymentLastStatus, error) {
+	database := db.client.Database(DatabaseName)
```
kjaskiewiczz, line 2782: in all other places we get the db name from context, which will result in the same name as you have here, but will look different in the enterprise, so I'm guessing there will be some conflicts?
sorry for the messy comment

docs/internal_api.yml
merlin-northern: discussed privately and changed.

store/mongo/datastore_mongo.go
merlin-northern: I assume we will migrate to single db one day, so thats why I am keeping it like that. do you want me to get the db from context?

merlin-northern:

docs/internal_api.yml
```diff
@@ -512,6 +512,43 @@ paths:
         500:
           $ref: "#/responses/InternalServerError"
 
+  /tenants/{tenant_id}/device/deployments/last:
```
kjaskiewiczz, line 515: since we accept list of ids, shouldn't the path contain /devices/ instead of /device/ ?

docs/internal_api.yml
```diff
@@ -512,6 +512,43 @@ paths:
...
+          in: path
+          description: Tenant identifier.
+          required: true
+          type: string
+        - name: devicesIds
+          in: body
+          description: An array of device ids for which to return the statuses.
+          type: array
+          items:
+            type: string
```
kjaskiewiczz, line 534:
```suggestion
          schema:
            type: array
            items:
              type: string
```

kjaskiewiczz: I'm also wondering if the list shouldn't be named (object with a list); this way it will be possible to extend it in the future if needed

docs/internal_api.yml
```diff
@@ -512,6 +512,43 @@ paths:
...
+          required: true
+      consumes:
+        - application/json
+      produces:
+        - application/json
+      responses:
+        200:
+          description: New configuration deployment created.
```
kjaskiewiczz, line 542: ??

kjaskiewiczz (changes requested):

docs/internal_api.yml
merlin-northern: sorry, copy-paste mistake.

docs/internal_api.yml
merlin-northern: fixed, together with schema and the wrapping object.

docs/internal_api.yml
merlin-northern: fixed. and the one below too.

api/http/api_deployments_last_status.go
```diff
@@ -0,0 +1,79 @@
...
+func (d *DeploymentsApiHandlers) GetDeviceDeploymentLastStatus(
+	w rest.ResponseWriter,
+	r *rest.Request,
+) {
+	l := requestlog.GetRequestLogger(r)
+
+	l.Debugf("starting")
+
+	tenantId := r.PathParam("tenant")
+	if tenantId == "" {
```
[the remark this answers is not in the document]
merlin-northern: removed the if. does it really hurt to add empty one? if it is empty we just match the empty string, which we also inserted in the db.save method, so maybe it is ok now?

docs/internal_api.yml
merlin-northern: fixed.

merlin-northern:
@mender-test-bot start pipeline -fast please

store/mongo/datastore_mongo.go
kjaskiewiczz: yes, like we do in the other places; this needs to work before migration too, and, despite the fact that we never use OS with multi tenancy, in theory, it's possible to use it this way

api/http/api_deployments_last_status.go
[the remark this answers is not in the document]
kjaskiewiczz: it doesn't make a difference, but it also doesn't make sense to add tenant to identity if the value is ""

api/http/api_deployments_last_status.go
```diff
@@ -0,0 +1,69 @@
...
+			l,
+			errors.Wrap(err, "cannot parse device ids array"),
+			http.StatusBadRequest,
+		)
+		return
+	}
+
+	l.Debugf("querying %d devices ids", len(devicesIds))
+	ctx := r.Context()
+	ctx = identity.WithContext(
```
kjaskiewiczz, line 54: I'd still appreciate if tenantId != "" here

docs/internal_api.yml
```diff
@@ -960,3 +992,15 @@ definitions:
...
+    properties:
+      device_deployment_last_statuses:
+        type: array
+        description: Array of device deployment statuses
+        items:
+          type: string
+    required:
+      - device_deployment_last_statuses
+    example:
+      device_deployment_last_statuses: "{\"device_deployment_last_statuses\":[{\"device_id\":\"b86dfe3d-a0a6-4838-b374-5fbcb7c956a1\",\"deployment_id\":\"acaf62f0-6a6f-45e4-9c52-838ee593cb62\",\"device_deployment_id\":\"b14a36d3-c1a9-408c-b128-bfb4808604f1\",\"device_deployment_status\":\"success\"}]}"
```
kjaskiewiczz, line 1006:
```suggestion
      device_deployment_last_statuses:
        - device_id: "b86dfe3d-a0a6-4838-b374-5fbcb7c956a1"
          deployment_id: "acaf62f0-6a6f-45e4-9c52-838ee593cb62"
          device_deployment_id: "b14a36d3-c1a9-408c-b128-bfb4808604f1"
          device_deployment_status: "success"

```

api/http/api_deployments_last_status.go
merlin-northern: sure. fixed.

docs/internal_api.yml
merlin-northern: sure fixed.

kjaskiewiczz (approved):
looks good

merlin-northern:
@mender-test-bot start pipeline --fast --pass please

PR #854, opened 2023-04-20
chore: Add `--throttle` flag to `propagate-reporting` command

alfrunes:


alfrunes:

main.go
```diff
@@ -91,6 +91,11 @@ func doMain(args []string) {
 					Name:  "tenant_id",
 					Usage: "Tenant ID (optional) - propagate for just a single tenant.",
 				},
+				cli.DurationFlag{
+					Name:  "throttle",
+					Usage: "`DURATION` to wait for each reindexing batch request",
```
merlin-northern, line 96: please sespecifycify what is a DURATION with units.

merlin-northern: also I find is slightly confusing name of the option is throttle and the usage says duration.

main.go
```diff
@@ -91,6 +91,11 @@ func doMain(args []string) {
...
+					Value: time.Millisecond * 25,
```
merlin-northern, line 97: could we have a const out of it? good news is: you can use it in addressing comments above :)

main.go
merlin-northern: I have not checked but will we have some examples of the command in the help? and: do we need to update the docs.mender.io?

propagete_reporting_test.go
```diff
@@ -131,7 +132,7 @@ func TestPropagateReporting(t *testing.T) {
 		t.Run(fmt.Sprintf("tc %s", k), func(t *testing.T) {
 			defer tc.workflowsMock.AssertExpectations(t)
 			defer tc.storeMock.AssertExpectations(t)
-			err := propagateReporting(tc.storeMock, tc.workflowsMock, tc.cmdTenant, tc.cmdDryRun)
+			err := propagateReporting(tc.storeMock, tc.workflowsMock, tc.cmdTenant, time.Microsecond, tc.cmdDryRun)
```
merlin-northern, line 135: we could create a new test case, which first calculates the avg time it took to run with throttle eq 0 and then sets the throttle to 2s and verifies that it took on avg longer, and the longer is propotional to the duration.

main.go
```diff
@@ -375,6 +385,8 @@ func reindexDeploymentsReporting(
 		if len(dd) < deviceDeploymentsBatchSize {
 			break
 		}
+		nextIterAt.Add(throttleDuration)
+		time.Sleep(time.Until(nextIterAt))
```
merlin-northern, line 389: I understand why you are doing it like that, but to be mean: what in case of the time change?

alfrunes: Right, I should use a timer to ensure that it's using monotonic time.

main.go
alfrunes: Changed the flat to `--rate-limit`. Should be more self-explanatory.

kjaskiewiczz (approved):

PR #860, opened 2023-05-03
chore: api docs check fix

merlin-northern:
Ticket: None


merlin-northern:
closing as the fix went in different way.

PR #862, opened 2023-05-04
chore: api docs check fix

merlin-northern:
api docs lint checks were not running due to:

WARNING: Entrypoint override disabled

and using the override itself.

example of a job in master:

https://gitlab.com/Northern.tech/Mender/deployments/-/jobs/4214764960

the problem with this fix: spectral lint returns 1 shall I add || true and if so, how to detect the failure?

Ticket: None


merlin-northern:
this is a proof of concept for https://github.com/mendersoftware/mendertesting/pull/253 it is very unfortunate that there is no way to test the changes to mendertesting, which impact all the pipelines.

merlin-northern:
we fixed this elsewhere (I believe in the runner config, it was long time already)

PR #863, opened 2023-05-09
fix: remove deprecated GET /releases endpoint

kjaskiewiczz:
Changelog: Title
Ticket: MEN-5180

BREAKING CHANGE: remove deprecated GET /releases endpoint


kjaskiewiczz:

merlin-northern (approved):
lgtm, as a review I checked if there are any leftovers; looks ok.
is the fact that we have `/deployments/releases/list` the reason for removing the endpoint in this PR? do we want to provide some fall-through to the former to preserve a bit of the API?

merlin-northern (approved):
lgtm, as a review I checked if there are any leftovers; found only [this](https://github.com/mendersoftware/deployments/pull/863#discussion_r1191955846). looks ok.
is the fact that we have `/deployments/releases/list` the reason for removing the endpoint in this PR? do we want to provide some fall-through to the former to preserve a bit of the API?

api/http/releases_test.go
merlin-northern, line 1: could you check `TestGetReleasesFilter` there is `reqUrl := "http://1.2.3.4/api/management/v1/deployments/releases"`

alfrunes (approved):

PR #865, opened 2023-05-19
Men 6474 skip verify flag tests

merlin-northern:


merlin-northern:
this was for testing.

PR #869, opened 2023-06-22
feat: New configuration parameter storage.proxy_uri

alfrunes:
The new configuration parameter rewrites presigned URLs to target the
specified proxy URL instead of the direct link. This allows setting
application level proxies in front of s3/azure blob storage.

Ticket: MEN-6562


alfrunes:

merlin-northern:
> he new configuration parameter rewrites presigned [..]

isnt configuration more of a she? ;-)

alfrunes:

storage/s3/options.go
```diff
@@ -44,26 +46,158 @@ var (
...
+func (s storageSettings) presignOptions(opts *s3.PresignOptions) {
+	if s.ExternalURI != nil {
+		presignURL := *s.ExternalURI
+		resolver := s3.EndpointResolverFromURL(presignURL,
+			func(ep *aws.Endpoint) {
+				ep.HostnameImmutable = s.ForcePathStyle
+			},
+		)
+		s3.WithPresignClientFromClientOptions(
+			s.options,
```
merlin-northern, line 138: this one is added (I am trying to prove that I really followed the moved code)

storage/s3/options.go
```diff
@@ -44,26 +46,158 @@ var (
...
+	}
+	if setting.ProxyURI != nil {
+		s.ProxyURI = setting.ProxyURI
+	}
+	if setting.ForcePathStyle != s.ForcePathStyle {
+		s.ForcePathStyle = setting.ForcePathStyle
+	}
+	if setting.UseAccelerate != s.UseAccelerate {
+		s.UseAccelerate = setting.UseAccelerate
+	}
```
merlin-northern, line 171: nit/question out of curiosity: why are we comparing here and for the rest we only check !=nil?

storage/s3/options.go
```diff
@@ -44,26 +46,158 @@ var (
...
+	if setting.ExternalURI != nil {
+		s.ExternalURI = setting.ExternalURI
+	}
+	if setting.URI != nil {
+		s.URI = setting.URI
+	}
+	if setting.ProxyURI != nil {
+		s.ProxyURI = setting.ProxyURI
+	}
```
merlin-northern, line 165: are we using pointer assignment on purpose here? or do we want a deeper copy?

storage/s3/options.go
```diff
@@ -44,26 +46,158 @@ var (
...
+	if setting.ForcePathStyle != s.ForcePathStyle {
+		s.ForcePathStyle = setting.ForcePathStyle
+	}
+	if setting.UseAccelerate != s.UseAccelerate {
+		s.UseAccelerate = setting.UseAccelerate
+	}
+	return s
+}
+
+func (s storageSettings) RewriteLink(link *model.Link) (*model.Link, error) {
```
merlin-northern, line 175: treat as idle comment and feel free to ignore: I am unsure if this should be the method of storageSetttings.

storage/s3/options.go
```diff
@@ -44,26 +46,158 @@ var (
...
+	return link, nil
+}
+
+type Options struct {
+	storageSettings
+
+	// ContentType of the uploaded objects
+	ContentType *string
+	// FilenameSuffix adds the suffix to the content-disposition for object downloads
+	FilenameSuffix *string
```
merlin-northern, line 200: not the easiest diff to read ;-)

storage/s3/options.go
```diff
@@ -87,30 +221,13 @@ func NewOptions(opts ...*Options) *Options {
 		BufferSize: &defaultBufferSize,
 	}
 	for _, opt := range opts {
-		if opt.StaticCredentials != nil {
-			ret.StaticCredentials = opt.StaticCredentials
-		}
-		if opt.Region != nil {
-			ret.Region = opt.Region
+		ret.storageSettings.patch(&opt.storageSettings)
```
merlin-northern, line 224:
```suggestion
		ret.storageSettings.patch(&opts[i].storageSettings)
```

storage/s3/options.go
```diff
@@ -87,30 +221,13 @@ func NewOptions(opts ...*Options) *Options {
 		BufferSize: &defaultBufferSize,
 	}
 	for _, opt := range opts {
```
merlin-northern, line 223:
```suggestion
	for i, opt := range opts {
```

storage/s3/options.go
```diff
@@ -87,30 +221,13 @@ func NewOptions(opts ...*Options) *Options {
...
-		if opt.StaticCredentials != nil {
-			ret.StaticCredentials = opt.StaticCredentials
-		}
-		if opt.Region != nil {
-			ret.Region = opt.Region
+		ret.storageSettings.patch(&opt.storageSettings)
+		if opt.DefaultExpire != nil {
+			ret.DefaultExpire = opt.DefaultExpire
```
merlin-northern, line 226: this is not in the `patch` on purpose, right?

storage/s3/options.go
```diff
@@ -290,18 +402,8 @@ func (opts *Options) toS3Options() (
...
 		s3.WithPresignExpires(expires)(s3Opts)
-		if opts.ExternalURI != nil {
-			presignURL := *opts.ExternalURI
-			resolver := s3.EndpointResolverFromURL(presignURL,
-				func(ep *aws.Endpoint) {
-					ep.HostnameImmutable = opts.ForcePathStyle
-				},
-			)
-			s3.WithPresignClientFromClientOptions(
-				s3.WithEndpointResolver(resolver),
```
merlin-northern, line 302: again showing that I am reading: presignOptions added `s.options` here

storage/s3/s3.go
```diff
@@ -165,18 +170,18 @@ func (s *SimpleStorageService) init(ctx context.Context) error {
 		case http.StatusForbidden:
 			err = fmt.Errorf(
 				"s3: insufficient permissions for accessing bucket '%s'",
-				s.bucket,
+				*s.settings.BucketName,
```
merlin-northern, line 173: are we sure it is !=nil here always?

storage/s3/options.go
alfrunes: It's a boolean

storage/s3/options.go
alfrunes: It's on purpose. It follows a similar option builder style as mongo/options.

storage/s3/options.go
alfrunes: The first commit did a refactoring, segregating the `model.StorageSettings` part of `Options` into an internal `storageSettings` type. Maybe it's a bit overkill and I should just keep the (bigger) `model.StorageSettings` for readability although it contains some irrelevant settings?

storage/s3/options.go
alfrunes: I'm not sure I understand the reasoning behind these two suggestions...

storage/s3/options.go
alfrunes: Patch applies to the `storageSettings` subset of the `Options`.

storage/s3/options.go
alfrunes: Yes, they were moved from inside `SimpleStorageService.*Request` interface to the `Options.preSignOptions` function to lower code repeatability.

storage/s3/s3.go
alfrunes: It should be non-empty from the `Options.Validate` call from `newClient`; however, this made me realize that this might break for the empty client, and thus when running with Azure as the main storage backend.

alfrunes: No, wait. We don't know :thinking:
Sorry about the mess.

alfrunes: Added a check at the start of the function

.gitlab-ci.yml
```diff
@@ -44,18 +44,4 @@ test:acceptance_tests:
...
-  image: golang:${GOLANG_VERSION}
-
-test:unit:
-  image: golang:1.18
-
-publish:acceptance:
-  image: golang:1.18-alpine3.16
-
-publish:unittests:
-  image: golang:1.18-alpine3.16
```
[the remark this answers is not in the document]
alfrunes: I removed all the image overrides, since these should come from the mendertesting template (which are also newer).

alfrunes:

alfrunes:

storage/azblob/azblob.go
```diff
@@ -423,6 +424,14 @@ func (c *client) GetRequest(
 			Reason:  err,
 		}
 	}
+	uri, err = applyProxyURL(uri, proxyURL)
```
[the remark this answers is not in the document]
alfrunes: Did some refactoring; moved to `utils` package and shared with `s3`.

storage/s3/options.go
alfrunes: Moved to `utils` and shared with `azblob`.

storage/s3/options.go
merlin-northern: I am paranoid, ever since I have realised that this:

```go
package main

import "fmt"

type Mode struct {
	number int
}

func main() {
	modes := []Mode{
		{
			number: 10,
		},
		{
			number: 11,
		},
	}
	for _, m := range modes {
		m.number = m.number * 2
	}
	fmt.Printf("modes:%+v", modes)
}
```

and this:
```go
package main

import "fmt"

type Mode struct {
	number int
}

func main() {
	modes := []Mode{
		{
			number: 10,
		},
		{
			number: 11,
		},
	}
	for i, _ := range modes {
		modes[i].number = modes[i].number * 2
	}
	fmt.Printf("modes:%+v", modes)
}
```

will actually print different results.

storage/azblob/azblob.go
```diff
@@ -450,9 +467,7 @@ func (c *client) DeleteRequest(
 			Reason:  err,
 		}
 	}
-	now := time.Now().UTC()
-	exp := now.Add(duration)
-	uri, err := bc.GetSASURL(sas.BlobPermissions{Delete: true}, now, exp)
```
merlin-northern, line 455: in both the delete and put requests we effectively removed that call, just to make me calm: why?

merlin-northern (approved):
left a couple of comments, thanks for being patient, pre-approving not to block. I know it was not the easiest patch.

storage/s3/options.go
alfrunes: That is correct, since the first example can be rewritten as:
```go
package main

import "fmt"

type Mode struct {
	number int
}

func main() {
	modes := []Mode{
		{
			number: 10,
		},
		{
			number: 11,
		},
	}
	for i := range modes {
		var m Mode = modes[i]
		m.number = m.number * 2
	}
	fmt.Printf("modes:%+v", modes)
}
```
However, since the iterable `opt` is read-only in this loop I find it cleaner to do this variable expansion inline.

storage/azblob/azblob.go
```diff
@@ -317,17 +320,60 @@ func (c *client) StatObject(
...
+	qParams, err := sas.BlobSignatureValues{
+		ContainerName: urlParts.ContainerName,
+		BlobName:      urlParts.BlobName,
+
+		Permissions:        permissions.String(),
+		ContentDisposition: contentDisposition,
+
+		StartTime:  now.UTC(),
+		ExpiryTime: exp.UTC(),
+	}.SignWithSharedKey(sk)
```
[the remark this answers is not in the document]
alfrunes: Yes, as you guessed, this is a "sort-of" reimplementation of the GetSASUrl. The reason is that the `azblob` API does not expose all the required parameters (we need to set the content disposition header for get requests). In this commit, I just collected this inside `buildURL` and shared the same path for the sake of re-using the same path.

storage/s3/s3.go
```diff
@@ -493,72 +504,86 @@ func (s *SimpleStorageService) PutObject(
...
-			opts,
+			opts.options,
 		)
 	} else if err == nil {
 		err = s.uploadMultipart(ctx, buf, path, src)
 	}
 	return err
 }
 
+func buildLink(
```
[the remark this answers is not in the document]
alfrunes: Let me see what I can do :+1:

PR #871, opened 2023-06-27
MEN-6348: Introduce release tags and endpoints for listing keys

alfrunes:
:information_source: Since there is no way of adding a limit to the projection scan (mongodb does not collapse $group and $limit to do partial index scans), I added internal limits on the number of unique tags that can be assigned to releases.


alfrunes:

alfrunes:
Converted to draft. I assumed that tags were key/value pairs as they are for inventory, but they are in fact just opaque strings.

alfrunes:
Superseded by #872

PR #882, opened 2023-07-06
docs: Document and improve message for 409 on artifact uploads

alfrunes:
Properly parses the server error from MongoDB and adds the conflicting dependency to the error response `metadata` attribute.

### DIFF
--- a/model/error.go
+++ b/model/error.go
@@ -1,4 +1,4 @@
-// Copyright 2020 Northern.tech AS
+// Copyright 2023 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


alfrunes:

kjaskiewiczz (commented):
it looks good, but it's way more than I expected from MEN-6444, and I have mixed feelings if this is good or bad ;)
the linter is confused and you need to update copyright dates;
when it comes to commit itself, with all your changes it's no longer "docs", it's more like fix or even feat - the error handling is not only fixed but extended

model/error.go
```diff
@@ -14,20 +14,49 @@
...
+	"encoding/json"
+	"errors"
+)
 
 type ConflictError struct {
-	Conflicts string
-	Err       error
+	Err       error       `json:"error"`
+	Metadata  interface{} `json:"metadata,omitempty"`
+	RequestID string      `json:"request_id,omitempty"`
```
mzedel, line 25: would it be possible to rely on [‎rest.utils/error.go](https://github.com/mendersoftware/go-lib-micro/blob/master/rest.utils/error.go#L17) or [rest_utils/api_error.go](https://github.com/mendersoftware/go-lib-micro/blob/master/rest_utils/api_error.go#L22) in these situations to be able to rely on a common error format? (maybe we could eventually even reference a common error definition in the api specs)

mzedel (commented):
in general LGTM and I'll make sure the extra info surfaces in the UI once this lands 👌

TheMeaningfulEngineer (approved):
I don't get the scope of the changes.
From my end, there was no 409 in the docs documented, now there is so that part is solved.

I see there is a new return object as well `ErrorExt`.
Assuming I'm reading it correctly it's under `/limits/storage`.
Don't have an opinion on that, haven't tried it.

model/error.go
alfrunes: The only issue is that the `rest[._]utils` errors does not wrap errors so the types would be in conflict. I guess I could make a new `ErrorV2` with proper initialization funcs. However, since this task originally was only 1SP, I'd like to defer that refactoring to some other time.

alfrunes:
> I don't get the scope of the changes. From my end, there was no 409 in the docs documented, now there is so that part is solved.
> 
> I see there is a new return object as well `ErrorExt`. Assuming I'm reading it correctly it's under `/limits/storage`. Don't have an opinion on that, haven't tried it.

Sorry, this PR became a bit more technical than the actual requirements for the task. The effect of this PR is that instead of just returning 409 with a message saying there was a conflict; it parses which `depends` key/value pair caused the conflict and returns this in a `metadata` object in the returned error. For example, say you upload an artifact with depends `{"foo": "bar"}`, you will now get an error on the form:
```json
{
  "error": "an artifact with the same name and depends already exists",
  "request_id": "59b1ce5e-a015-4bcf-84bf-80892e93c67c",
  "metadata": {
    "conflicts": {
      "foo": "bar"
    }
  }
}
```

kjaskiewiczz (approved):

TheMeaningfulEngineer (approved):

PR #883, opened 2023-07-10
fix: Delete device deployment history incorrectly triggering reindex

alfrunes:
The endpoint was missing the reporting feature flag check, which caused it to send reindexing request even though the reporting service is disabled.


alfrunes:

kjaskiewiczz (approved):

PR #884, opened 2023-07-12
feat: release notes: update and store.

merlin-northern:
Changelog: Title
Ticket: MEN-6593


merlin-northern:
tested manually. working on tests.

merlin-northern:
@alfrunes I am trying to come up with acceptance tests, otherwise ready for review.

model/release.go
```diff
@@ -146,11 +146,49 @@ func (tag *Tag) UnmarshalJSON(b []byte) error {
...
+func (n Notes) Validate() error {
+	length := len(n)
+	if length > NotesLengthMaximumCharacters {
+		return ErrReleaseNotesTooLong
+	}
+	for i := 0; i < length; i++ {
+		if !allowedByte(n[i]) {
+			return ErrCharactersNotAllowed
+		}
+	}
```
alfrunes, line 173: You could use `unicode.IsGraphic` to check that the text is printable. As for XSS, the json library escapes HTML by default when marshalling. You could alternatively make `json.Unmarshaler` that calls `json.HTMLEscape` on the decoded string if you're paranoid of storing strings without escaping them first :sweat_smile: 
```go
// Error type is optional, but it helps the end-user to understand what is not allowed.
type InvalidCharError struct {
	Offset int
	Char   rune
}

func (err *CharError) Error() string {
	return fmt.Sprintf(`invalid character '%c' at offset %d`, err.Char, err.Offset)
} 
```
```suggestion
	if i := strings.IndexFunc(string(n), unicode.IsPrint); i > 0 {
		return InvalidCharError{
			Char: n[i],
			Offset: i,
		}
	}
```

store/mongo/datastore_mongo_releases.go
```diff
@@ -0,0 +1,209 @@
...
+				Value: bson.D{
+					{
+						Key: StorageKeyReleaseNotes, Value: release.Notes,
+					},
+				},
+			},
+		},
+	)
+	if err != nil {
+		return errors.WithMessage(err, "mongo: failed to update release tags")
```
alfrunes, line 204:
```suggestion
		return errors.WithMessage(err, "mongo: failed to update release")
```

api/http/routing.go
```diff
@@ -63,7 +63,8 @@ const (
 	ApiUrlManagementLimitsName = ApiUrlManagement + "/limits/#name"
 
 	ApiUrlManagementV2               = "/api/management/v2/deployments"
-	ApiUrlManagementV2ReleaseTags    = ApiUrlManagementV2 + "/releases/#name/tags"
+	ApiUrlManagementV2ReleasesName   = ApiUrlManagementV2 + "/deployments/releases/#name"
+	ApiUrlManagementV2ReleaseTags    = ApiUrlManagementV2ReleasesName + "/tags"
```
alfrunes, line 67: Not sure if a `PUT /releases/#name/notes` would be more appropriate than an incomplete `PATCH /releases/#name`?

alfrunes (commented):
Found a couple of remarks here and there.
FYI: Moving code around makes it easier to miss things in reviews.

api/http/routing.go
merlin-northern: I prepared it so we can easily in the future update other fields, so I kindly suggest to keep it as is.

store/mongo/datastore_mongo_releases.go
merlin-northern: copy-paste error. fixed. thanks.

merlin-northern:
> Found a couple of remarks here and there. FYI: Moving code around makes it easier to miss things in reviews.

thank you. sometimes you have to move some code. sorry for the inconvenience.

merlin-northern:
@mender-test-bot start pipeline --fast please

model/release.go
```diff
@@ -146,11 +147,57 @@ func (tag *Tag) UnmarshalJSON(b []byte) error {
...
+	ErrReleaseNotesTooLong  = errors.New("release notes too long")
+	ErrCharactersNotAllowed = errors.New("release notes contain characters which are not allowed")
+)
+
+type InvalidCharError struct {
+	Offset int
+	Char   byte
+}
+
+func (err InvalidCharError) Error() string {
```
alfrunes, line 164: (nitpick) Generally, it's preferred to have pointer receiver for errors. This way `errors.Is` compares by pointer reference instead of struct header.
```suggestion
func (err *InvalidCharError) Error() string {
```

model/release.go
```diff
@@ -146,11 +147,57 @@ func (tag *Tag) UnmarshalJSON(b []byte) error {
...
+	return !unicode.IsGraphic(r)
+}
+
+func (n Notes) Validate() error {
+	length := len(n)
+	if length > NotesLengthMaximumCharacters {
+		return ErrReleaseNotesTooLong
+	}
+	if i := strings.IndexFunc(string(n), IsNotGraphic); i > 0 {
+		return InvalidCharError{
```
alfrunes, line 178:
```suggestion
		return &InvalidCharError{
```

tests/common.py
```diff
@@ -36,6 +37,33 @@
...
+
+    def unlock(self):
+        if not os.path.exists(MONGO_LOCK_FILE):
+            return
+        with open(MONGO_LOCK_FILE, "r") as fh:
+            hash = fh.read()
+            if hash != self.hash:
+                return
+        os.remove(MONGO_LOCK_FILE)
+
```
alfrunes, line 66: I'm not sure if I get this. The acceptance test has to run serially are you using xdist locally?

alfrunes: Class name does indicate what kind of lock this is. Consider renaming it to `MongoLock`.

model/release.go
merlin-northern: sure, no problem.

model/release.go
merlin-northern: no problema (I went the other way and removed the * from the receiver)

tests/common.py
merlin-northern: you know, they are using the same db, and I have seen strange things. so I figured there should be some locking.
I thought of giving Lock() a parameter in the constructor, perhaps you like it?

merlin-northern:
@mender-test-bot start pipeline --fast please

alfrunes (approved):

merlin-northern:
@mender-test-bot start pipeline --fast please baby

merlin-northern:
last push was a squash and distribution fo recent changes between commits. [0 diff](https://github.com/mendersoftware/deployments/compare/1d2cd9b7fca37038829c0821570323f0893553df..f725351ef9d214080ef5994a068b6a6d87cfe272)

PR #901, opened 2023-08-18
feat: save and get the update types

merlin-northern:
Changelog: Title
Ticket: MEN-6623


store/mongo/datastore_mongo_releases.go
```diff
@@ -219,3 +220,60 @@ func (db *DataStoreMongo) UpdateRelease(
...
+
+	tenantId := ""
+	if id := identity.FromContext(ctx); id != nil {
+		tenantId = id.Tenant
+	}
+	result := c.FindOne(
+		ctx,
+		bson.M{
+			StorageKeyId:       StorageKeyStorageReleaseUpdateTypes,
+			StorageKeyTenantId: tenantId,
```
merlin-northern, line 266: shall I create an index for that?

app/app.go
```diff
@@ -412,7 +413,27 @@ func (d *Deployments) handleArtifact(ctx context.Context,
...
+	}
+	if i > 0 {
+		err = d.db.SaveUpdateTypes(ctx, updateTypes[:i])
+		if err != nil {
+			l.Errorf(
+				"error while saving the update types for the artifact: %s",
+				err.Error(),
+			)
+		}
+	}
```
tranchitella, line 436:
```suggestion
	var updateTypes []string
	if image.ArtifactMeta != nil && len(image.ArtifactMeta.Updates) > 0 {
		updateTypes = make([]string, len(image.ArtifactMeta.Updates))
		for _, t := range image.ArtifactMeta.Updates {
			if t.TypeInfo.Type == nil {
				continue
			}
			updateTypes[i] = *t.TypeInfo.Type
		}
		if err = d.db.SaveUpdateTypes(ctx, updateTypes[:i]); err != nil {
			l.Errorf(
				"error while saving the update types for the artifact: %s",
				err.Error(),
			)
		}
	}
```

store/mongo/datastore_mongo_releases.go
tranchitella: Why not using the tenantId as `_id`? This way you don't need to create a new index.
The `StorageKeyStorageReleaseUpdateTypes` as _id is useless.

tranchitella (commented):
First round of review

store/mongo/datastore_mongo_releases.go
merlin-northern: > Why not using the tenantId as `_id`? This way you don't need to create a new index. The `StorageKeyStorageReleaseUpdateTypes` as _id is useless.

I did not like the idea that tenant_id it can be empty. the `StorageKeyStorageReleaseUpdateTypes` is not that bad, mongo does use this schema to refer to "singular objects in a collection". anyways, if you want me to move to tenant_id what do we do with the case when tenant_id is empty?

app/app.go
merlin-northern: sure, moved the second block into the if.

merlin-northern:
@tranchitella addressed the comments (I will squash to one commit before merging): https://github.com/mendersoftware/deployments/compare/30d2e92bac6b0ab6473600f2fa67853d68faaa8d..295a349aad496ef02b53de677b3aca6cb6596fa6

if you insist on not using `StorageKeyStorageReleaseUpdateTypes` as an id for the singular document, let's talk, but I thought it maybe a good idea (mongo uses this schema:
```
> use admin
switched to db admin
> db.system.version.find()
{ "_id" : "featureCompatibilityVersion", "version" : "4.2" }
```
)
otherwise I will just add the index and all the dance.

store/mongo/datastore_mongo_releases.go
tranchitella: Ok, makes sense.

tranchitella:
> @tranchitella addressed the comments (I will squash to one commit before merging): https://github.com/mendersoftware/deployments/compare/30d2e92bac6b0ab6473600f2fa67853d68faaa8d..295a349aad496ef02b53de677b3aca6cb6596fa6
> 
> if you insist on not using `StorageKeyStorageReleaseUpdateTypes` as an id for the singular document, let's talk, but I thought it maybe a good idea (mongo uses this schema:
> 
> ```
> > use admin
> switched to db admin
> > db.system.version.find()
> { "_id" : "featureCompatibilityVersion", "version" : "4.2" }
> ```
> 
> ) otherwise I will just add the index and all the dance.

Up to you, I'm okay with the current implementation but you need to add an index.

tranchitella (approved):

store/mongo/migration_1_2_18.go
```diff
@@ -0,0 +1,57 @@
...
+
+// Up creates an index for filtering tags and retrieving distinct tag keys.
+func (m *migration_1_2_18) Up(from migrate.Version) error {
+	ctx := context.Background()
+	idxReleases := m.client.
+		Database(m.db).
+		Collection(CollectionUpdateTypes).
+		Indexes()
+
+	_, err := idxReleases.CreateOne(ctx, mongo.IndexModel{
```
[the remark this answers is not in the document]
merlin-northern: you are right. I will further simplify.

merlin-northern:
@mender-test-bot start pipeline  --fast please sugar

docs/management_api_v2.yml
```diff
@@ -178,6 +201,13 @@ definitions:
     items:
       type: string
 
+  UpdateTypes:
+    type: array
+    description: |-
+      Update types as present in the images.
+    items:
+      type: string
+
```
merlin-northern, line 210: @tranchitella added this one -- for completness

store/mongo/datastore_mongo.go
```diff
@@ -472,6 +476,8 @@ const (
 	StorageKeyStorageSettingsForcePathStyle = "force_path_style"
 	StorageKeyStorageSettingsUseAccelerate  = "use_accelerate"
 
+	StorageKeyStorageReleaseUpdateTypes = "update_types"
+
```
merlin-northern, line 480: @tranchitella index and migration with tests are added

store/mongo/datastore_mongo_releases.go
```diff
@@ -219,3 +220,62 @@ func (db *DataStoreMongo) UpdateRelease(
...
+	tenantId := ""
+	if id := identity.FromContext(ctx); id != nil {
+		tenantId = id.Tenant
+	}
+	options := mopts.UpdateOptions{}
+	options.SetUpsert(true)
+	_, err := c.UpdateOne(
+		ctx,
+		bson.M{
+			StorageKeyTenantId: tenantId,
```
merlin-northern, line 242: I got rid fo the _id as it is useless -- the index ensure the uniqueness

tranchitella (approved):

PR #914, opened 2023-09-05
Fix/api docs

mzedel:


docs/management_api.yml
```diff
@@ -2248,6 +2253,12 @@ definitions:
           letters, digits, underscores, periods and hyphens.
         items:
           type: string
+      notes:
+        type: string
+        description: |
+          Additional information describing the content of a Release or a set of Artifacts
+          The notes are limited to some maximum length and we won't tell you! HARHAR! just kidding it's 1024 characters 👌
+          also please use the v2 API to set this... just because. (same with the tags, but it's written here to not be too obvious :-*)
```
mzedel, line 2261: @merlin-northern am I doing this right?

merlin-northern, line 2261: syntax seems correct. We tend to be more formal in the docs, so I would suggest (kindly) to rephrase a bit.
```suggestion
          Additional information describing a Release limited to 1024 bytes. Please use v2 API to set this field. 
```

mzedel: Under intense pressure I applied your suggestion 👌

merlin-northern: you must be getting old, this is it, right? ;-)

merlin-northern (approved):

docs/management_api.yml
merlin-northern: thank you for fixing that!

mzedel:

PR #916, opened 2023-09-06
fix: fixed an issue that would cause a malformed response if no update types exist yet

mzedel:
Ticket: None
Changelog: None

### DIFF
--- a/store/mongo/datastore_mongo_releases.go
+++ b/store/mongo/datastore_mongo_releases.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/pkg/errors"
 	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
 	mopts "go.mongodb.org/mongo-driver/mongo/options"
 
 	"github.com/mendersoftware/go-lib-micro/identity"
@@ -279,6 +280,9 @@ func (db *DataStoreMongo) GetUpdateTypes(ctx context.Context) ([]string, error) 
 	}
 	var updateTypes updateType
 	err := result.Decode(&updateTypes)
+	if err == mongo.ErrNoDocuments {
+		return []string{}, nil
+	}
 	if err != nil {
 		return []string{}, err
 	} else {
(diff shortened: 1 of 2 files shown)


store/mongo/datastore_mongo_releases.go
```diff
@@ -279,6 +280,9 @@ func (db *DataStoreMongo) GetUpdateTypes(ctx context.Context) ([]string, error)
 	}
 	var updateTypes updateType
 	err := result.Decode(&updateTypes)
+	if err == mongo.ErrNoDocuments {
+		return []string{}, nil
+	}
```
merlin-northern, line 285: so you want no error on no documents. ok.

tests/tests/test_release_v2.py
```diff
@@ -142,3 +142,24 @@ def test_get_all_releases_types(self, mongo, cli):
...
+                + f"/api/management/v2/deployments/releases/all/types"
+            )
+
+            r = requests.get(
+                get_types_url,
+                verify=False,
+                headers={"Authorization": "Bearer foo"},
+            )
+            types = json.loads(r.text)
+            assert types == []
```
merlin-northern, line 165: we could probably assert here that there was no error (HTTP code success)

merlin-northern (approved):
misc comment in the tests only. thank you!

store/mongo/datastore_mongo_releases.go
mzedel: not quite... the API spec promises an array, as the endpoint is supposed to respond with an array and an empty array is a valid response if no items to populate the list exist...

What I _want_ is for the backend to validate the backend endpoints with the API specs.

tests/tests/test_release_v2.py
mzedel: sure, but I wanted to align with the test right above (where I took this from).

mzedel:

store/mongo/datastore_mongo_releases.go
mzedel: no: https://github.com/mendersoftware/deployments/blob/master/api/http/api_deployments_releases.go#L192-L196

merlin-northern: ok then!

PR #917, opened 2023-09-07
chore: align api docs with implementation

mzedel:


mzedel:

merlin-northern (approved):
thanks for fixing -- we are not parsing _v2 docs spec in the tests, the decision was to wait until we port it to openapi.

PR #923, opened 2023-09-24
feat: filter the list of releases by tags

tranchitella:
Add support for filtering the release list by tags; this commit deprecates the end-point `/v1/deployments/releases/lists` and introduces a new end-point `/v2/deployments/releases/list` to fix the capitalization of the keys in the response payload, which was capitalized by mistake and cannot be changed without breaking API back-compatibility.

Ticket: MEN-6349
Changelog: deprecate the `/v1/deployments/releases/list` end-point, replaced by the new end-point `/v2/deployments/releases`
Changelog: add support for filtering releases by tags


tranchitella:

tranchitella:

docs/management_api_v2.yml
```diff
@@ -44,6 +44,182 @@ responses:
...
+      description: |
+        Returns a collection of releases, allows filtering by release name and sorting
+        by name or last modification date.
+      parameters:
+        - name: name
+          in: query
+          description: Release name filter.
+          required: false
+          type: string
+        - name: device_type
```
kjaskiewiczz, line 65: do we still want to filter by update type? is it included in the new UX for releases?

docs/management_api_v2.yml
```diff
@@ -44,6 +44,182 @@ responses:
...
+        - ManagementJWT: []
+      summary: |
+        List releases
+      description: |
+        Returns a collection of releases, allows filtering by release name and sorting
+        by name or last modification date.
+      parameters:
+        - name: name
+          in: query
+          description: Release name filter.
```
kjaskiewiczz, line 62: Are we planning to support search for name as it was in the old endpoint, or just name prefix as in the new UX design?

store/mongo/datastore_mongo.go
```diff
@@ -767,16 +767,31 @@ func (db *DataStoreMongo) getReleases_1_2_15(
...
 		if filt.DeviceType != "" {
-			filter[StorageKeyReleaseArtifactsDeviceTypes] = bson.M{"$regex": filt.DeviceType}
+			filter[StorageKeyReleaseArtifactsDeviceTypes] = bson.M{"$regex": primitive.Regex{
+				Pattern: ".*" + regexp.QuoteMeta(filt.DeviceType) + ".*",
+				Options: "i",
+			}}
 		}
 		if filt.UpdateType != "" {
-			filter[StorageKeyReleaseArtifactsUpdateTypes] = bson.M{"$eq": filt.UpdateType}
+			filter[StorageKeyReleaseArtifactsUpdateTypes] = bson.M{"$regex": primitive.Regex{
```
kjaskiewiczz, line 791: do we need to support regular expression for update types or just exact match?

kjaskiewiczz (commented):
do we have indexes for fields we're using for filtering and sorting?

docs/management_api_v2.yml
mzedel: update type yes ([see here](https://www.figma.com/file/xHtArTgENL7BZHGBNYX1Vr/Delta-artifacts?node-id=809%3A1296&mode=dev))
OTOH `device_type` is not in the UX, but not sure if this is needed elsewhere...

kjaskiewiczz: right, sorry; I meant `device_type`

api/http/api_deployments.go
merlin-northern, line 1: small typo in the commit message: "response paylod" -> "response payload"

merlin-northern (commented):
I think we are testing only the deprecated endpoint in the acceptance tests. cold you also fix that? (this is my fault) (it is possible that I am talking about a different endpoint, so just verify if we are testing properly)

api/http/api_deployments.go
tranchitella: Fixed the typo in the commit message; we do support total count and links, it was already implemented.

docs/management_api_v2.yml
tranchitella: implemented name prefix

docs/management_api_v2.yml
tranchitella: to me looks kind of "obvious" to provide a filter by device_type, I'd keep it at least in the API.
The cost of this extra filter is minimal. Do you disagree, @kjaskiewiczz (and @alfrunes, @merlin-northern, @mzedel)?

tranchitella:
> do we have indexes for fields we're using for filtering and sorting?

Yes, we do. However, they won't be used in combination and there is not too much we can do.

docs/management_api_v2.yml
tranchitella: EDIT: changed my mind, let's drop it.

docs/management_api.yml
```diff
@@ -886,10 +886,9 @@ paths:
         Returns a collection of releases, allows filtering by release name and sorting
         by name or last modification date.
 
-        DEPRECATED: this end-point is deprecated because there is a mismatch in the
-        capitalization of the fields of the response body and it doesn't support
-        advanced filters and sorting, please use the v2 /deployments/releases
-        end-point instead.
+        DEPRECATED: due to a mismatch in the  capitalization of the fields of the response
```
merlin-northern, line 889:
```suggestion
        DEPRECATED: due to a mismatch in the capitalization of the fields of the response
```

merlin-northern (approved):
lgtm

tranchitella:

kjaskiewiczz (approved):
looks good

PR #936, opened 2023-10-04
fix: db.Find: look for phrases (escape the $text elements)

merlin-northern:
We have to escape the $text search to look for a phrase, otherwise there are special characters, and individual terms searching in place.

Changelog: Title
Ticket: MEN-6730


merlin-northern:

tranchitella (approved):

PR #939, opened 2023-10-12
Fix: Make comment more readable for users 

MuchoLucho:
that may end up troubleshooting to this point


MuchoLucho:
This suggestion came from a support ticket from an enterprise customer. They ended up troubleshooting in the code and this message did not provide any insights.
Eventually when we found the root cause, we would like to make it clearer.

tranchitella:
Thanks @MuchoLucho , add this to your commit message and force-push:

```
Changelog: None
Ticket: None
```

MuchoLucho:
Hello @tranchitella, ready. Thanks for the follow up

PR #948, opened 2023-10-30
Upgrade go mod dependencies

alfrunes:


alfrunes:

alfrunes:
Changing the driver to use the `BaseEndpoint` has a breaking change to the interface. Apparently it is impossible to use transfer acceleration with a custom endpoint (which is why the test fails). I'm tempted to change the config default for AWS_URI to blank string and let the resolver assemble the URL.

alfrunes:

tranchitella (approved):

PR #959, opened 2023-11-08
chore: enable api docs linter as obligatory

merlin-northern:
Ticket: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -63,3 +63,6 @@ publish:image:mender:
 # FIXME Revert the image overrides once templates in mendertesting is updated
 test:static:
   image: golangci/golangci-lint:v1.53.3
+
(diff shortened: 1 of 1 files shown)


mzedel (approved):

merlin-northern:
@lluiscampos is that the correct way to enable api linter as obligatory for specific repos?

merlin-northern:
@mzedel wow! faster review ever :)

lluiscampos:
> @lluiscampos is that the correct way to enable api linter as obligatory for specific repos?

It will work, yes :+1: 

This is backend team responsibility, I don't want to overreach here. How I would do it:
* Set it to `false` in the template (or remove it, which is the default)
* Explicitly allow failures in the repos that are not yet fixed
* Have tasks in the backlog for all these repos

PR #960, opened 2023-11-10
refac: change the way we check if artifact is part of any deployment

kjaskiewiczz:


kjaskiewiczz:

PR #961, opened 2023-11-10
add support for bulk removal of releases by names

kjaskiewiczz:
Ticket: [MEN-6354](https://northerntech.atlassian.net/browse/MEN-6354)


docs/management_api_v2.yml
```diff
@@ -215,6 +215,42 @@ paths:
...
+        - name: delete_request
+          in: body
+          required: true
+          type: object
+          properties:
+            artifact_names:
+              type: array
+              description: List of names of artifacts to delete.
+              items:
+                type: string
```
mzedel, line 241:
```suggestion
      parameters:
        - name: delete_request
          in: body
          required: true
          type: object
          properties:
            release_names:
              type: array
              description: List of names of releases to delete.
              items:
                type: string
```

to stick to the release focus or maybe rather:

```
      parameters:
        - name: name
          in: query
          description: Name of the release to be deleted
          required: true
          type: string
```

to align with the existing filtering structure when listing releases (so `DELETE /deployments/releases?name=foo&name=bar` along with `GET /deployments/releases?name=baz&name=icecream`)

OTOH you likely know other reasons not to use the query params for that 😅

kjaskiewiczz: I'm fine with using query params here; thanks

kjaskiewiczz:

store/datastore.go
```diff
@@ -49,7 +49,7 @@ type DataStore interface {
 	//tenants
 	ProvisionTenant(ctx context.Context, tenantId string) error
 
-	//images
+	// images
```
tranchitella, line 52: really stupid thing: why are you adding a space here? All other comments don't have spaces :)

store/mongo/datastore_mongo.go
```diff
@@ -2959,6 +2939,48 @@ func (db *DataStoreMongo) UpdateDeploymentsWithArtifactName(
...
+	if err != nil {
+		return []string{}, err
+	}
+	defer cursor.Close(ctx)
+
+	var deployments []*model.Deployment
+	if err = cursor.All(ctx, &deployments); err != nil {
+		if err == mongo.ErrNoDocuments {
+			return []string{}, nil
+		}
```
tranchitella, line 2972:
```suggestion
		if err == mongo.ErrNoDocuments {
			err = nil
		}
```

tranchitella (approved):

store/datastore.go
kjaskiewiczz: It's leftover from some other work, which I didn't include in this PR at the end.
I was moving things around and, by accident, I removed this line. Then I added it back, but with space.
I think it was me who put the comments without spaces in the first place, but I think they look better with spaces.
Do you want me to remove the space?

store/mongo/datastore_mongo.go
kjaskiewiczz: done

kjaskiewiczz:

PR #988, opened 2024-01-10
MEN-6916

tranchitella:


docs/devices_api.yml
```diff
@@ -39,8 +39,174 @@ responses:
     description: Invalid Request.
     schema:
       $ref: "#/definitions/Error"
+  UnauthorizedError: # 401
+    description: Unauthorized.
+    schema:
+      $ref: "#/definitions/Error"
 
 paths:
+  /artifacts:
```
merlin-northern, line 48: lets introduce the limit of max releases and document it here. the pagination we can add later without breaking this API.

docs/devices_api.yml
```diff
@@ -39,8 +39,174 @@ responses:
...
+    get:
+      operationId: List Artifacts
+      tags:
+        - Device API
+      security:
+        - DeviceJWT: []
+      summary: List artifacts filtering by release and device_type
```
merlin-northern, line 55:
```suggestion
      summary: List artifacts filtering by release name and device type
```

merlin-northern (commented):
one general question: how are we going to get the URLs to the actual files? the same way as /next?

tranchitella:
> one general question: how are we going to get the URLs to the actual files? the same way as /next?

Yes.

docs/devices_api.yml
merlin-northern: copy that. but I believe it should be documented.

kjaskiewiczz (approved):
looks good;

PR #1012, opened 2024-04-05
feat: Add configuration for max data size when generating artifacts

alfrunes:
Adds a new configuration option for setting the max data section size when generating an image with a default of 512MiB. The configuraiton path is `storage.max_generate_data_size` or environment variable `DEPLOYMENTS_STORAGE_MAX_GENERATE_DATA_SIZE`.

### DIFF
--- a/api/http/api_deployments.go
+++ b/api/http/api_deployments.go
@@ -52,8 +52,9 @@ const (
 	// 15 minutes
 	DefaultDownloadLinkExpire = 15 * time.Minute
 	// 10 Mb
-	DefaultMaxMetaSize  = 1024 * 1024 * 10
-	DefaultMaxImageSize = 10 * 1024 * 1024 * 1024 // 10GiB
+	DefaultMaxMetaSize         = 1024 * 1024 * 10
+	DefaultMaxImageSize        = 10 * 1024 * 1024 * 1024 // 10GiB
+	DefaultMaxGenerateDataSize = 512 * 1024 * 1024       // 512MiB
 
 	// Pagination
 	DefaultPerPage                      = 20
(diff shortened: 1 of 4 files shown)


alfrunes:

api/http/api_deployments.go
```diff
@@ -52,8 +52,9 @@ const (
 	// 15 minutes
 	DefaultDownloadLinkExpire = 15 * time.Minute
 	// 10 Mb
-	DefaultMaxMetaSize  = 1024 * 1024 * 10
-	DefaultMaxImageSize = 10 * 1024 * 1024 * 1024 // 10GiB
+	DefaultMaxMetaSize         = 1024 * 1024 * 10
+	DefaultMaxImageSize        = 10 * 1024 * 1024 * 1024 // 10GiB
+	DefaultMaxGenerateDataSize = 512 * 1024 * 1024       // 512MiB
```
merlin-northern, line 57: normally I would say: we should indicate that it is bytes in the identifier, but I see we dont do it here.

api/http/api_deployments.go
```diff
@@ -976,7 +987,7 @@ ParseLoop:
 			}
 			// Add one since this will impose the upper limit on the
 			// artifact size.
-			if size > d.config.MaxImageSize {
```
merlin-northern, line 979: just to double check, for what are we now using `MaxImageSize`?
also I do not understand the comment with "Add one"

merlin-northern (approved):
looks good to me, but KJ is the expert on deployments.

api/http/api_deployments.go
alfrunes: > also I do not understand the comment with "Add one"

I have no idea. It seems like this may have been a comment for a different if-branch from the past? :man_shrugging:

PR #1013, opened 2024-04-08
fix: Improve error message when uploading too large artifacts

alfrunes:
Discovered this confusing error message while testing #1012


alfrunes:

PR #1018, opened 2024-05-06
feat: prevent the creation of deployments if there is already an active deployment with the same constructor parameters

tranchitella:


tranchitella:

kjaskiewiczz (commented):
~~I think we need to update API docs too~~
EDIT: sorry, I only briefly looked at the PR and thought we'll return 409 (which is undocumented), but we're returning 400;
I don't feel like 400 Bad Request is the best error code in this case.
IMO 409 will do better (some people are using 303, which feels a bit exotic?)
github is returning 422 in similar cases, and for me 422 was always a return code indicating that I'm giving up on http rfc ;) (still I'm using it from time to time)
@alfrunes @merlin-northern any opinions?

tranchitella:
> ~I think we need to update API docs too~ EDIT: sorry, I only briefly looked at the PR and thought we'll return 409 (which is undocumented), but we're returning 400; I don't feel like 400 Bad Request is the best error code in this case. IMO 409 will do better (some people are using 303, which feels a bit exotic?) github is returning 422 in similar cases, and for me 422 was always a return code indicating that I'm giving up on http rfc ;) (still I'm using it from time to time) @alfrunes @merlin-northern any opinions?

look at the latest commit

docs/management_api.yml
```diff
@@ -38,6 +38,10 @@ responses:
     description: Invalid Request.
     schema:
       $ref: "#/definitions/Error"
+  ConflictError: # 409
+    description: Conflict
```
kjaskiewiczz, line 42: What about being more specific here?
Like `An active deployment with the exact same parameters already exists`?

kjaskiewiczz (approved):
Looks good to me, but maybe it's a good idea to wait for @alfrunes or @merlin-northern opinion on the status code we're returning here?

docs/management_api.yml
tranchitella: Updated

alfrunes (approved):

tranchitella:

PR #1021, opened 2024-05-08
Acceptance test fixup

alfrunes:
- [X] Replace minio with boto3 (S3) library
- [x] Use pytest semantics for loading coftest parameters into test fixtures
  - Gave up on using fixtures since the setup is too convoluted and relying on just-in-time loading of options. 
- [x] Remove hard-coded values for mongo URL and docker compose labels in `CliClient`


alfrunes:

tests/run.sh
```diff
@@ -17,6 +17,8 @@
 
 DIR=$(readlink -f $(dirname $0))
 
+pip install boto3 # FIXME
+
```
tranchitella, line 21: This is not great; however, I understand it is required to make it work.

tranchitella (approved):
I approve; however, the `boto3 installation` as part of `tests/run.sh` should not be there and I highly suggest updating the upstream docker image instead.

tests/run.sh
alfrunes: Sorry, I forgot to put link in the PR, but I created one here: https://github.com/mendersoftware/mender-test-containers/pull/569
I will remove the top commit before I merge (once the mender-test-container is published).

PR #1023, opened 2024-05-27
fix release filtering and RBAC for releases

kjaskiewiczz:
With the old implementation, when using more than one tag in the filter, or when using role which grants access to releases with given tag (and more than one tag was specified), deployments will present only releases containg ALL the tags each.
With the new behavior, deployments will retrun all the releases containg ANY of the tags.

Ticket: MEN-7272


kjaskiewiczz:

tranchitella (approved):

PR #1026, opened 2024-06-11
perf: Optimize lookup for the next deployment by narrowing search

alfrunes:


store/mongo/datastore_mongo.go
```diff
@@ -2760,6 +2763,41 @@ func (db *DataStoreMongo) FindNewerActiveDeployments(ctx context.Context,
 	return deployments, nil
 }
 
+// FindNewerActiveDeployment finds active deployments which were created
+// after createdAfter for which is either dynamic (requires post processing)
```
kjaskiewiczz, line 2767: We do not support dynamic deployments here (the code is OK, it's only the comment which is wrong).

store/mongo/datastore_mongo.go
```diff
@@ -463,6 +463,8 @@ const (
 	StorageKeyDeploymentActive              = "active"
 	StorageKeyDeploymentStatus              = "status"
 	StorageKeyDeploymentCreated             = "created"
+	StorageKeyDeploymentDeviceList          = "device_list"
+	StorageKeyDeploymentDynamic             = "dynamic"
```
kjaskiewiczz, line 467: no dynamic deployments here

kjaskiewiczz (changes requested):
There is no support for dynamic deployments here, so please update the code and commit message. Except for that it's looking good.

alfrunes:
> There is no support for dynamic deployments here, so please update the code and commit message. Except for that it's looking good.

Right, thanks. I actually just backported the commit from enterprise. I'll trim the dynamic deployments references from this PR.

alfrunes:

kjaskiewiczz (approved):

PR #1029, opened 2024-06-27
perf: Optimize database interactions when updating deployment stats

alfrunes:
Changed the sequential find and double update ops with a findAndModify followed by a conditional update if the status actually changed. This also improves atomicity as there is an inherent race condition when doing the deployment status aggregation locally.

I also noticed that most internal calls to UpdateDeviceDeploymentStatus already has the DeviceDeployment already, so I added an internal updateDeviceDeploymentStatus which skips fetching the device deployment (again).

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1397,20 +1397,28 @@ func (d *Deployments) createDeviceDeploymentWithStatus(
 		return nil, err
 	}
 
-	// after inserting new device deployment update deployment stats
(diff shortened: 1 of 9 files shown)


alfrunes:

kjaskiewiczz (approved):
looks good

tranchitella (approved):
Nice optimization!

PR #1036, opened 2024-08-13
fix: excessive patterns in regexp for image and releases search removed

merlin-northern:
Changelog: Title
Ticket: None

### DIFF
--- a/store/mongo/datastore_mongo.go
+++ b/store/mongo/datastore_mongo.go
@@ -560,7 +560,7 @@ func (db *DataStoreMongo) getReleases_1_2_14(
 			{Key: "$match", Value: bson.M{
 				StorageKeyImageName: bson.M{
 					"$regex": primitive.Regex{
-						Pattern: ".*" + regexp.QuoteMeta(filt.Name) + ".*",
+						Pattern: "^" + regexp.QuoteMeta(filt.Name),
 						Options: "i",
 					},
 				},
@@ -593,7 +593,7 @@ func (db *DataStoreMongo) getReleases_1_2_14(
 			{Key: "$match", Value: bson.M{
 				"artifacts." + StorageKeyImageDescription: bson.M{
 					"$regex": primitive.Regex{
-						Pattern: ".*" + regexp.QuoteMeta(filt.Description) + ".*",
+						Pattern: regexp.QuoteMeta(filt.Description),
 						Options: "i",
 					},
 				},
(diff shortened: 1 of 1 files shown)


merlin-northern:

merlin-northern:
I do not really think it works with any regular expressions more complicated that just a string to look for. so either we should remove regexp as it is expensive and move to just looking for a sub string, or fix it to work properly.

merlin-northern:

store/mongo/datastore_mongo.go
```diff
@@ -560,7 +560,7 @@ func (db *DataStoreMongo) getReleases_1_2_14(
 			{Key: "$match", Value: bson.M{
 				StorageKeyImageName: bson.M{
 					"$regex": primitive.Regex{
-						Pattern: ".*" + regexp.QuoteMeta(filt.Name) + ".*",
+						Pattern: regexp.QuoteMeta(filt.Name),
```
alfrunes, line 563: In the UI it says "filter by prefix". Shouldn't it rather be like this?
```suggestion
						Pattern: "^" + regexp.QuoteMeta(filt.Name),
```

merlin-northern: I know, and I do not have a problem with changing it, although it will break the undocumented and probably unused API. @kjaskiewiczz opinions?

alfrunes: IMO it is fixing a bug. We're saying it is prefix search, but it is returning documents that simply contain the query as a substring. This is leading to confusing/inconsistent results.

kjaskiewiczz: the UI says "starts with", the API was wrongly documented, so I guess we can fix it by searching for prefix

merlin-northern: the task which this is a spinoff from was about clarifying this behaviour, and thats what I did, described how it works now. I can fix this PR and the docs PR, no problem.

merlin-northern: okey then. let me align both of the PRs.

tranchitella: I agree, I would fix it making it work as a prefix search.

merlin-northern: okey so as we all agree. I adjusted:
1. name is searched always by prefix
2. description and type searched anywhere (I honestly do not see UI using it at all)

merlin-northern:
@mender-test-bot start pipeline --fast please

alfrunes (approved):

kjaskiewiczz (approved):

PR #1045, opened 2024-11-14
chore(deps): update docker image to golang:1.23.3-bullseye

kjaskiewiczz:


tests/tests/test_configuration_deployments.py
```diff
@@ -322,7 +324,8 @@ def verify_artifact(self, fname, name, dtype, config):
         assert "Type: mender-configure" in stdout
         assert "Name: {}".format(name) in stdout
         assert "Version: 3" in stdout
-        assert "Compatible devices: '[{}]'".format(dtype) in stdout
+        # NOTE: Using regular expression for backward compatibility
+        assert re.search(rf"Compatible devices: (\[{dtype}\]|'\[{dtype}\]')", stdout)
```
merlin-northern, line 328: we have to create mender-artifact smoke test to ensure that the output does not change. the utility jak mernder-artifact is used in the field in scripts, we cannot modify the output as we please.

merlin-northern (approved):

PR #1048, opened 2025-02-13
feat: enable exact artifact name, description, and device type matching in list artifacts requests

Michaelhobo:
When using the List Artifacts APIs to find artifacts by name, the
existing implementation only supports $regex search using the pattern
".*<name>.*". This is CPU intensive on large image collections, when the
user only needs an exact match. On an existing dataset with 60000
images, the existing $regex query took 421ms while the new exact_name
query took 9ms. This was benchmarked using both device_types and name
set. Also, the new exact match makes an index more effective.

Adds exact_name, exact_device_type, and exact_description boolean params
to the GET /artifacts and GET /artifacts/list APIs. For example,
/artifacts/list?name=123.t would match name="123.test" but
/artifacts/list?name=123.t&exact_name would not match.

Tested: Ran a local deployments instance hooked up to a dev mongodb
atlas instance. Used curl with exact_name and exact_device_type, tested
that inexact matches were not returned using the new exact_ params but
were returned without the exact_ params.

### DIFF
--- a/api/http/api_deployments.go
+++ b/api/http/api_deployments.go
@@ -71,19 +71,22 @@ const (
 // storage keys
 const (
 	// Common HTTP form parameters
-	ParamArtifactName = "artifact_name"
-	ParamDeviceType   = "device_type"
-	ParamUpdateType   = "update_type"
-	ParamDeploymentID = "deployment_id"
-	ParamDeviceID     = "device_id"
-	ParamTenantID     = "tenant_id"
-	ParamName         = "name"
-	ParamTag          = "tag"
-	ParamDescription  = "description"
-	ParamPage         = "page"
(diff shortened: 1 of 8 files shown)


kjaskiewiczz:
@mho-nuro thank you for your contribution.
Your PR is not passing our tests, but there are only small things to fix - unit tests and commit linting.
For the unit tests - I'm not sure why you modified test for deleting images, but it's failing now.
For commit formatting - we are using conventional commits, so it would be great if you could add "feat: " prefix to your commit title.

Michaelhobo:
@kjaskiewiczz I've fixed the checks

kjaskiewiczz (approved):

